PR #13, opened 2022-11-02
Develop hubbard correction

ndaelman-hu:
Addition of DFT+U metadata (called `HubbardModel`) and parsing for VASP and FHI-aims.
On the GUI side, is `U_effective` shown in the DFT search menu, while `U` and `J` are just indexed in elasticsearch.
These features were first discussed in https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/issues/884


ladinesa:
Can you also include the pylint fixes for the wannier90 parser here?

ndaelman-hu:
@JosePizarro3 has already amended it in the QMC branch.
I'm not sure when it will be merged, though.

electronicparsers/vasp/parser.py
```diff
@@ -1175,7 +1176,23 @@ def parse_method(self):
...
+        # atom properties & hubbard correction
+        # based on https://www.vasp.at/wiki/index.php/LDAUTYPE (08/07/2022)
+        hubbard_method_type = {1: 'Liechtenstein', 2: 'Dudarev', 4: 'Liechtenstein without exchange splitting'}
+        hubbard_orbital_type = {-1: None, 0: 's', 1: 'p', 2: 'd', 3: 'f'}
+        hubbard_present = False
+
+        # try parsing vasprun.xml, incar, outcar (abbreviated settings)
+        for file_type, parser_type in zip(['incar_out', 'incar', 'incar', 'incar_out'],
+                                          [RunContentParser, RunContentParser, OutcarContentParser, OutcarContentParser]):
+            self.parser.incar
```
ladinesa: not sure what this line should do.

tests/test_vaspparser.py
```diff
@@ -226,3 +226,30 @@ def test_metagga(parser):
...
+    archive = EntryArchive()
+    # test the values in vasprun.xml and INCAR (selected in place of OUTCAR)
+    reference_dir = 'tests/data/vasp/Mg4V2Bi2O12_dftu/'
+    for mainfile in ['vasprun.xml', 'OUTCAR']:
+        parser.parse(reference_dir + mainfile, archive, None)
+        params = archive.run[-1].method[-1].atom_parameters
+        references = {'orbital': ('d', 'd'), 'u': (3.25, 2.), 'j': (0., 1.)}
+        slice = 0
+        for param in params:
+            if 'hubbard_model' in param.__dict__.keys():
```
ladinesa: if param.hubbard_model:

electronicparsers/vasp/parser.py
ndaelman-hu: So, the OUTCAR abbreviates all U and J definitions to 2-point decimals.
If we are reading a run.xml file, we can go with the usual routine of first checking that and then falling back on the INCAR.
But for Hubbard parameters, we should invert the order.

ladinesa: I mean line 1188, it is just a an expression.

ndaelman-hu: For some kind of reason, the `incar` variables don't seem to be fully instantiated, and I will get errors trying to access them.
Once I run them, everything works.
Amir ran into something similar in a GUI feature and this was the best solution there.

ladinesa: You get error when you do `self.parser.incar[filte_type]` in 1189? I think we should investigate `def incar`. I think we should not access private `_incar`

electronicparsers/vasp/parser.py
```diff
@@ -1175,7 +1176,23 @@ def parse_method(self):
                 except Exception:
                     self.logger.warn('Error setting metainfo', data=dict(key=key))
 
-        # atom properties
+        # atom properties & hubbard correction
+        # based on https://www.vasp.at/wiki/index.php/LDAUTYPE (08/07/2022)
+        hubbard_method_type = {1: 'Liechtenstein', 2: 'Dudarev', 4: 'Liechtenstein without exchange splitting'}
```
ladinesa: a matter of taste but I think we should put this in the class init as e.g. self._hubbard_method_types and probably merging it with orbital type?

electronicparsers/fhiaims/parser.py
```diff
@@ -1349,6 +1357,13 @@ def parse_atom_type(species):
                     sec_atom_species.x_fhi_aims_controlInOut_species_cut_pot = val[0][0] * ureg.angstrom
                     sec_atom_species.x_fhi_aims_controlInOut_species_cut_pot_width = val[0][1] * ureg.angstrom
                     sec_atom_species.x_fhi_aims_controlInOut_species_cut_pot_scale = val[0][2]
+                elif "request for '+U'" in key:
+                    sec_hubbard = sec_atom_type.m_create(HubbardModel)
+                    sec_hubbard.orbital = str(val[0][0]) + val[0][1]
```
ladinesa: I prefer` f'{val[0][0]}{val[0][1]'}`

tests/test_vaspparser.py
```diff
@@ -226,3 +226,30 @@ def test_metagga(parser):
...
+                assert param.hubbard_model.method == 'Dudarev'
+                assert param.hubbard_model.orbital == references['orbital'][slice]
+                assert approx(param.hubbard_model.u.to(ureg.eV).magnitude) == references['u'][slice]
+                assert approx(param.hubbard_model.j.to(ureg.eV).magnitude) == references['j'][slice]
+                slice += 1
+
+    # check the OUTCAR value when no INCAR is present
+    parser.parse('tests/data/vasp/Mg4V2Bi2O12_dftu_no_incar/OUTCAR', archive, None)
+    params = archive.run[-1].method[-1].atom_parameters
+    assert approx(params[1].hubbard_model.u.to(ureg.eV).magnitude) == 3.2
```
ladinesa, line 255: no approx error here?

ndaelman-hu: No, since it is only defined up to 1 decimal point.
I can add them if you want.

electronicparsers/vasp/parser.py
ndaelman-hu: You mean into the **method** `VASPParser.__init__()`, right?
I'm all fine with that, but I don't understand _merging with orbital types_.

ladinesa: Yes. I don' know if I make sense but something like `{1: ('Liechtenstein', 's)...`. So we only have one mapping.

ndaelman-hu: No, it doesn't: these are 2 completely unrelated parameters, i.e. the type of Hubbard method used and to which band it is applied.
I'll push as you suggested, but with 2 separate maps.

electronicparsers/vasp/parser.py
ndaelman-hu: Using `self.parser.incar` causes OUTCAR to overwrite the INCAR, which is what we were trying to avoid for Hubbard parameters.
In general, this behavior makes sense, though.

ndaelman-hu: I'm unable atm to reproduce the error when leaving line 1188 out.
Should I

- [ ] leave it in and try reproducing later?
- [x] comment it out, so we still know the workaround later?
- [ ] leave it out and only revisit it once we start seeing errors again?

ndaelman-hu:
Am getting the following pylint error:
`************* Module electronicparsers.fhiaims.metainfo.fhi_aims
electronicparsers/fhiaims/metainfo/fhi_aims.py:1431:19: E1101: Module 'nomad.datamodel.metainfo.simulation.method' has no 'HubbardModel' member (no-member)`

Locally, I cannot reproduce it though.
Instead I get another error for flpo, since its dependency `DFTPlusU` was changed. Should I for the time being just update the name? Later on, I can consider this parser more thoroughly.

ladinesa:
This is the error that I am referring to that is why I do not want to merge to develop because the nomad pypi does not yet include the new metainfo defs. But anyway, let us run the tests again when the pakage is released. For now as long as tests run fine locally I think we are ok.

ndaelman-hu:
Okay, now I see! Should I still push the minor updates to the fplo parser?
Else, I fear that it might break.

ladinesa:
The fplo stuff is still related to hubbard? If so then add it as long as it passes your local tests.

ndaelman-hu:
I was using extending the old class DFTPlusU, now renamed to HubbardModel (which is more appropriated).
There are no tests built for DFT+U in fplo, but it doesn't seem to have been implemented in the parser anyhow, just set up in the metadata...

ladinesa:
Then I think you need to update every parser to use HubbardModel. Just create an issue to add tests later.

ndaelman-hu:
Fplo was the only one left

PR #15, opened 2022-11-09
> Fix Fermi level regex in different versions of FHIaims

JosePizarro3:
@ndaelman-hu @ladinesa I implemented the regex for the two fermi_level quantities (depending on the version).

Please, feel free to test it (mainly @ndaelman-hu) and to improve the regexes (mainly @ladinesa).

### DIFF
--- a/electronicparsers/fhiaims/parser.py
+++ b/electronicparsers/fhiaims/parser.py
@@ -336,8 +336,13 @@ class FHIAimsOutParser(TextParser):
                 'lumo', r'Lowest unoccupied state \(CBM\) at\s*([\d\.\-\+Ee ]+) (?P<__unit>\w+)',
                 repeats=False, dtype=float),
             Quantity(
-                'fermi_level', rf'{re_n} *\| Chemical potential \(Fermi level\) in (?P<__unit>\w+)\s*:([\d\.\-\+Ee ]+)',
-                repeats=False, dtype=float)]
+                'fermi_level',  # older version
+                rf'{re_n} *\| Chemical potential \(Fermi level\) in (\w+)\s*:([\d\.\-\+Ee ]+)',
+                str_operation=lambda x: float(x.split()[1]) * units_mapping.get(x.split()[0])),
+            Quantity(
+                'fermi_level',  # newer version
+                rf'{re_n} *\| Chemical potential \(Fermi level\)\:\s*([\-\d\.]+)\s*(\w+)',
+                str_operation=lambda x: float(x.split()[0]) * units_mapping.get(x.split()[1], 1))]
 
         def str_to_scf_convergence2(val_in):
             val = val_in.split('|')
@@ -943,6 +948,10 @@ class FHIAimsParser:
 
             if iteration.get('fermi_level') is not None:
                 sec_energy.fermi = iteration.get('fermi_level')
+                try:
+                    iteration.get('fermi_level').units
+                except Exception:
+                    self.logger.warn('Erorr setting the Fermi level: no units')
 
             # eigenvalues scf iteration
             eigenvalues = get_eigenvalues(iteration)
--- a/tests/test_fhiaimsparser.py
+++ b/tests/test_fhiaimsparser.py
@@ -124,6 +124,10 @@ def test_band_spinpol(parser):
     assert sec_dos.total[0].value[46].magnitude == approx(1.2418253e-11)
     assert sec_dos.total[1].value[15].magnitude == approx(3.91517047e-11)
 
+    # v151211 test for the Fermi level
+    assert sec_scc.energy.fermi.to('eV').magnitude == approx(-9.3842209)
+    assert sec_k_band.energy_fermi == sec_scc.energy.fermi
+
 
 def test_band_silicon(silicon):
     """Tests that the band structure of silicon is parsed correctly.
@@ -146,6 +150,10 @@ def test_band_silicon(silicon):
     highest_occupied_index = lowest_unoccupied_index - 1
     gap = energies[lowest_unoccupied_index] - energies[highest_occupied_index]
     assert gap == approx(0.60684)
+
+    # v071914_7 test for the Fermi level
+    assert scc.energy.fermi.to('eV').magnitude == approx(-5.7308573)
+    assert band.energy_fermi == scc.energy.fermi
 
 
 def test_dos_silicon(silicon):


ndaelman-hu:
@ladinesa Do the "missing" metadata in the parsers mean that the copy of the nomad `develop` branch has to be updated here?

ndaelman-hu:
Okay, so firstly, let me link this to the corresponding Gitlab issues, namely https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/issues/939 and https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/issues/1131

@JosePizarro3 I didn't find any update to the FHI tester. One of us should update them. Note that the current test is very subtle, but didn't pick up on the issue. In `test_fhiaimsparser.py` there are 3 places where the Fermi energy is called: twice in Si_band_dos (v071914_7) and once in Fe_band_spinpol (v151211). It is not directly tested in the old version (which you corrected), only the band gap is. So, please add tests for `Si_band_dos`, similar to `Fe_band_spinpol` and maybe use comments to specify that two different versions / formats are being tested.

Secondly -and this is my own pet peeve (@ladinesa feel free to comment)- the reference data is too unclear, as it is referenced in SI units. I suggest to, henceforth, always put the reference data in the same units as reported and convert the extracted data. You can look at `test_dftu()` for an example. This will greatly improve casual lookups.

ladinesa:
If you are refering to the no-member error due to the missing metainfo defs like HubbardModel, we simply have to wait until a new nomad pypi is released then this will be automatically resolved. We do not need to update the parsers, we just rerun the actions. For the value comparisons, I did the opposite before but I agree that it is easier to look at to compare with the raw value and convert the nomad value. Can you please do this to all tests?

JosePizarro3:
I added some tests now in `test_band_spinpol` and `test_band_silicon`.

ndaelman-hu:
@JosePizarro3 Thx! If passes your pytest locally, then I'm fine.
@ladinesa I will progressively implement the change in the future.

electronicparsers/fhiaims/parser.py
```diff
@@ -336,8 +336,13 @@ def str_to_species(val_in):
                 'lumo', r'Lowest unoccupied state \(CBM\) at\s*([\d\.\-\+Ee ]+) (?P<__unit>\w+)',
                 repeats=False, dtype=float),
             Quantity(
-                'fermi_level', rf'{re_n} *\| Chemical potential \(Fermi level\) in (?P<__unit>\w+)\s*:([\d\.\-\+Ee ]+)',
-                repeats=False, dtype=float)]
+                'fermi_level',
+                rf'{re_n} *\| Chemical potential \(Fermi level\) in (\w+)\s*:([\d\.\-\+Ee ]+)',
```
ladinesa, line 340: doesn't re_float work here?

electronicparsers/fhiaims/parser.py
```diff
@@ -336,8 +336,13 @@ def str_to_species(val_in):
...
+                str_operation=lambda x: float(x.split(' ')[2]) * units_mapping.get(x.split(' ')[0])),
+            Quantity(
+                'fermi_level',
+                rf'{re_n} *\| Chemical potential \(Fermi level\)\:\s*([\-\d\.]+)\s*(\w+)',
+                str_operation=lambda x: float(x.split(' ')[0]) * units_mapping.get(x.split(' ')[1]))]
```
ladinesa: x.split() should just work, you actually did it twice, if you use an external function not lambda, you can probably do it once only, but I am fine with this.

ladinesa: You can also set the default unit to 1. i.e. units_mapping.get(x.split()[1], 1) just in case there exists no unit. Or you deliberately cause a error here but you should be able to handle the error in writing the data to archive, e.g. put an exception 'Error settting fermi'

electronicparsers/fhiaims/parser.py
```diff
@@ -336,8 +336,13 @@ def str_to_species(val_in):
                 'lumo', r'Lowest unoccupied state \(CBM\) at\s*([\d\.\-\+Ee ]+) (?P<__unit>\w+)',
                 repeats=False, dtype=float),
             Quantity(
-                'fermi_level', rf'{re_n} *\| Chemical potential \(Fermi level\) in (?P<__unit>\w+)\s*:([\d\.\-\+Ee ]+)',
-                repeats=False, dtype=float)]
```
ndaelman-hu, line 340: Please add an inline comment stating which regex is _old-style_ and which is _new-style_

tests/test_fhiaimsparser.py
```diff
@@ -124,6 +124,10 @@ def test_band_spinpol(parser):
     assert sec_dos.total[0].value[46].magnitude == approx(1.2418253e-11)
     assert sec_dos.total[1].value[15].magnitude == approx(3.91517047e-11)
 
+    # v151211 test for the Fermi level
```
ndaelman-hu, line 127: I made a mistake in the comments: v151211 still follows _old-style_
I've attached one of my own test folders, please add this one to test _new-style_ and rename its _old-style_ counterpart to the same convention

[Si_band_dos_v171221_1.zip](https://github.com/nomad-coe/electronic-parsers/files/9970507/Si_band_dos_v171221_1.zip)

ndaelman-hu (changes requested):

electronicparsers/fhiaims/parser.py
JosePizarro3: It does not :/

ladinesa: I think we should update re_float such that the digit before the decimal point becomes optional i.e. `re_float = r'[-+]?\d*\.\d*(?:[Ee][-+]\d+)?'`. not sure though if this will mess up the others. otherwise I am fine without using re_float here.

electronicparsers/fhiaims/parser.py
JosePizarro3: Check out the exception I wrote.

electronicparsers/fhiaims/parser.py
JosePizarro3: Done

tests/test_fhiaimsparser.py
JosePizarro3: We conveyed you will do it. I have push the latest changes addressing your comments.

electronicparsers/fhiaims/parser.py
JosePizarro3: I'd prefer not to touch re_float, and use the regex we know it works already.

ndaelman-hu (approved):

JosePizarro3:
@ladinesa Nathan and I talked and he will push his changes later on. If you give your final approval I will merge.

ladinesa (approved):

PR #24, opened 2022-12-01
Implement new workflow normalization

ladinesa:


ndaelman-hu:
## Investigation Results

I've surveyed all parsers that set `intensities_units` and verified the physical quantities:

### Parsers Setting intensities_units

1. **Exciting** (`electronicparsers/exciting/parser.py:2333`)
   - Sets: `'F/m'`
   - Calculates: Imaginary part of macroscopic dielectric function (ε₂)
   - **Actual units**: Dimensionless ❌

2. **OCEAN** (`electronicparsers/ocean/parser.py:239`)
   - Sets: `'F/m'`  
   - Calculates: Absorption/emission spectra via BSE (also ε₂)
   - **Actual units**: Dimensionless ❌

### Physical Background

Both parsers calculate optical spectra using the Bethe-Salpeter Equation (BSE). The output is the imaginary part of the macroscopic dielectric function ε₂(ω), which is **dimensionless**.

As confirmed by GPAW, ABINIT, VASP, and BerkeleyGW documentation: the dielectric function (both ε₁ and ε₂) is a dimensionless quantity.

### Proposed Schema Convention

For consistency with NOMAD's existing convention (e.g., `magnetic_shielding` uses `unit='dimensionless'`), I propose:

**Set `intensities_units = 'dimensionless'` for dimensionless spectra**

This is clearer than empty string or omitting the field, and aligns with how dimensionless Quantity fields are marked throughout NOMAD schemas.

PR #26, opened 2022-12-02
Delete parse_gw outdated method from parse_configurations

JosePizarro3:
I restored the FHIaims parser deleting these extra outdated lines. After talking with Adam, they were probably regenerated in a rebasing.


lauri-codes (approved):
Removes an obsolete function

PR #28, opened 2022-12-15
Wann90 changes

JosePizarro3:
I want to improve a bit the parser to make the method `parse_system` callable.

This is because many advanced many-body methods (like DMFT) don't store anywhere the system geometry, except for from the Wannier90 output files themselves. This way I can call `Wannier90Parser:parse_system` in more advanced parser.

### DIFF
--- a/electronicparsers/wannier90/parser.py
+++ b/electronicparsers/wannier90/parser.py
@@ -60,12 +60,6 @@ class WOutParser(TextParser):
             Quantity(
                 'positions', rf'\|\s*([\-\d\.]+)\s*([\-\d\.]+)\s*([\-\d\.]+)', repeats=True)]
 
-        orbitals_quantities = [
-            Quantity(
-                'names', r'\|\s*([A-Za-z\.\s\-\/]+)\|', repeats=False),
-            Quantity(
-                'parameters', r'\|\s*([\d\.\s\-\/]*)\|', repeats=True)]
-
         self._quantities = [
             # Program quantities
             Quantity(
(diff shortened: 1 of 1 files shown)


electronicparsers/wannier90/parser.py
```diff
@@ -140,21 +150,60 @@ def __init__(self):
...
+            'sp3d-3': [-4, 3],
+            'sp3d-4': [-4, 4],
+            'sp3d-5': [-4, 5],
+            'sp3d2-1': [-5, 1],
+            'sp3d2-2': [-5, 2],
+            'sp3d2-3': [-5, 3],
+            'sp3d2-4': [-5, 4],
+            'sp3d2-5': [-5, 5],
+            'sp3d2-6': [-5, 6]
+        }
```
lauri-codes: You have added this variable, but none of your new code uses it?

lauri-codes (commented):
Done with the review here. Is the pipeline error expected?

JosePizarro3:
> Done with the review here. Is the pipeline error expected?

Shouldn't be, as that metainfo is defined in nomad. @ndaelman-hu @ladinesa do you have any idea of why pylint is not passing here?

ladinesa:
This is expected as the target nomad is v1.0.8

JosePizarro3:
Should we update it now to a higher, e.g. 1.1.5? Or something is preventing us of doing it so?

And also, this is just about modifying the yaml in .github/workflows? I am curious.

ladinesa:
Yes, but the package should be released which I think is not.

JosePizarro3:
Ok, then this is ready for merging. I deleted the extra unused lines, so I merge directly.

PR #29, opened 2022-12-21
w2dynamics DMFT parser

JosePizarro3:
I created the parser for w2dynamics, as well as changes in datamodel, normalizers, and gui of nomad: see https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/issues/1156

I am not sure if the way of parsing the system is the best. I do it from the Wannier90 files. I can speak with the devs so they output this data in the w2dynamics output files directly.

To be done: modify the parser to input any tight-binding model, and not only Wannier90.

### DIFF
--- a/electronicparsers/w2dynamics/parser.py
+++ b/electronicparsers/w2dynamics/parser.py
@@ -21,8 +21,8 @@ import os
 import logging
 import h5py
 import re
-from nomad.units import ureg
 
+from nomad.units import ureg
 from nomad.parsing.file_parser import TextParser, Quantity
 from nomad.datamodel.metainfo.simulation.run import Run, Program
 from nomad.datamodel.metainfo.simulation.calculation import (
--- a/electronicparsers/w2dynamics/parser.py
+++ b/electronicparsers/w2dynamics/parser.py
@@ -121,7 +121,7 @@ class W2DynamicsParser:
                 setattr(target, f'x_w2dynamics_{name}', value[:])
             # mu is a single value
             if key == 'mu':
-                setattr(target, f'x_w2dynamics_{name}', value)
+                target.x_w2dynamics_mu = value
 
     def parse_system(self):
         wann90_files = [f for f in os.listdir(self.maindir) if f.endswith('.wout')]
--- a/electronicparsers/w2dynamics/parser.py
+++ b/electronicparsers/w2dynamics/parser.py
@@ -1,0 +1,285 @@
+#
+# Copyright The NOMAD Authors.
+#
+# This file is part of NOMAD.
+# See https://nomad-lab.eu for further info.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


JosePizarro3:
Hi @ladinesa ,

Can you please review this parser?

electronicparsers/w2dynamics/parser.py
```diff
@@ -0,0 +1,285 @@
...
+from nomad.datamodel.metainfo.simulation.method import (
+    Method, HoppingMatrix, HubbardKanamoriModel, LatticeModelHamiltonian, DMFT
+)
+from nomad.datamodel.metainfo.workflow import Workflow
+from .metainfo.w2dynamics import (
+    x_w2dynamics_axes, x_w2dynamics_quantities, x_w2dynamics_config_parameters,
+    x_w2dynamics_config_atoms_parameters, x_w2dynamics_config_general_parameters,
+    x_w2dynamics_config_qmc_parameters
+)
+from ..wannier90.parser import Wannier90Parser, WOutParser, HrParser
```
ladinesa, line 40: Maybe try to put them in a common class.

electronicparsers/w2dynamics/parser.py
```diff
@@ -0,0 +1,285 @@
...
+            value = source[key]
+            if isinstance(value, h5py.Group) and 'value' in value.keys():
+                value = value['value']
+            if not isinstance(value, h5py.Dataset):
+                continue
+            name = self._re_namesafe.sub('_', key)
+            if value.shape:
+                setattr(target, f'x_w2dynamics_{name}', value[:])
+            # mu is a single value
+            if key == 'mu':
```
ladinesa, line 123: can simply be x_w2dynamics_mu = value

electronicparsers/w2dynamics/parser.py
ladinesa: Disregard commnet, so there are really wannier files.

ladinesa (approved):
Very nice. please merge.

electronicparsers/w2dynamics/parser.py
JosePizarro3: I think this is the most open question right now. I guess we have 3 possible solutions (even separable in two categories: short and long term); feel free to correct me if I am wrong:

1. Leave it as it is right, i.e., parsing certain sections from another parser if the files are present. This is the shortest term patch I can imagine to.
2. Automatic workflow: with this, one could link between DFT>Wannier90>w2dynamics, but this requires largest mods and sitting together (note there are several DFT (e.g. VASP, QE, wien2k...), Projection (Wannier90, Tight Binding Studio...), and DMFT (w2dynamics, TRIQS...) codes, so this automatization should recognize the correct parser for each upload). There is also a few risks, starting with "what happens if the files for all these 3 tasks are in different folders, or they are uploaded at different times in different entries".
3. User-defined workflow from a json / yaml file: this is the safest, but it requires educating the users to know how to define these workflows. For me, this is the longest term solution, but it might be the best one.

We could discuss this more in depth next year with Area C/D, as it will define also the future of workflows 🙂

electronicparsers/w2dynamics/parser.py
JosePizarro3: You meant:

`if key == 'mu':`
`     target.x_w2dynamics_mu = value`

?

JosePizarro3: ok, I changed to my comment, resolve it if this is what you meant :)

PR #30, opened 2022-12-21
Mod in Wannier90 method.projection

JosePizarro3:
I restructured the Method.Projection section to contain 2 subsections: Wannier, which is the one defined currently in nomad, and SlaterKoster, which will be defined soon.

This is because the projection can be done with two distinct methodologies (the above ones). Later I will push a parser for a code which deals with the SlaterKoster-like projection.

### DIFF
--- a/electronicparsers/wannier90/parser.py
+++ b/electronicparsers/wannier90/parser.py
@@ -27,7 +27,7 @@ from nomad.datamodel.metainfo.simulation.run import Run, Program
 from nomad.datamodel.metainfo.simulation.calculation import (
     Calculation, Dos, DosValues, BandStructure, BandEnergies, Energy, HoppingMatrix
 )
-from nomad.datamodel.metainfo.simulation.method import Method, KMesh, Projection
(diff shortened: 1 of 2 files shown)


JosePizarro3:
Hi @ladinesa ,

Two questions:

1. Can you quickly review these changes? 
2. In general, when I have such small changes or perhaps hotfixes in the submodule, can I simply merge them myself or do you want me to still contact you for reviewing?

ladinesa:
You can simply merge them I alreadt lifted the review requirement

ladinesa:
But please always run a local test to make sure all tests pass

ladinesa (approved):

ladinesa (approved):

JosePizarro3:
Ok, I checked the tests and of course one was failing. Now it is fixed.

I also took the chance to fix the `parse_system`. I saw an upload on Nomad that was not taking correctly the system because there were multiple regexes for the lattice_vectors. Now it works.

I merge this then.

PR #33, opened 2023-01-04
w2dynamics parser changes

JosePizarro3:
I changed the metainfo in nomad, hence I have to modify the w2dynamics parser. This was done to generalize further the DMFT metainfo.


JosePizarro3:
@ladinesa the local testing passes, so I merge directly this small changes

PR #37, opened 2023-01-16
HubbardModel to HubbardKanamoriModel renaming

JosePizarro3:
Please, @ndaelman-hu , this is quickly to review, can you take a look?

Some remarks for the metainfo:

- `orbital` is now MEnum, accepting the orbital angular momentum label, without number.
- ~~electron_volts everywhere~~
- `method` renamed to `double_counting_correction`
- `projection_type` moved as code-specific quantities

Please, also [check the MR in nomad regarding this](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/issues/1285). I will link it with your [current issue on DFT+U](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/issues/884).

### DIFF
--- a/electronicparsers/fhiaims/metainfo/fhi_aims.py
+++ b/electronicparsers/fhiaims/metainfo/fhi_aims.py
@@ -1441,9 +1441,16 @@ class AtomParameters(simulation.method.AtomParameters):
         repeats=True)
 
 
-class HubbardModel(simulation.method.HubbardModel):
+class HubbardKanamoriModel(simulation.method.HubbardKanamoriModel):
 
     m_def = Section(validate=False, extends_base_section=True)
+
+    x_fhi_aims_projection_type = Quantity(
+        type=str,
+        shape=[],
+        description='''
+        Type of orbitals used for projection in order to calculate occupation numbers.
+        ''')
 
     x_fhi_aims_petukhov_mixing_factor = Quantity(
         type=np.dtype(np.float64),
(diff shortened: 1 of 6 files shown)


ndaelman-hu:
> electron_volts everywhere

What do you mean here? The units depend on the parser (indeed eV for VASP and FHIaims), while NOMAD stores the values as joule.

JosePizarro3:
It means `u_effective` is in electron_volts in nomad, indeed. That's why I got rid of the `ureg.eV` units in here.

ndaelman-hu:
> It means u_effective is in electron_volts in nomad, indeed. That's why I got rid of the ureg.eV units in here.

Yeah, I saw it from your implementation now. To my understanding, this runs against the SI convention.
Did you verify with Luca or anyone from Area D whether we're allowed to do this?
Don't forget that when querying via the API, the user only gets numbers, not quantities. If we change units like this, the queried data becomes meaningless...

JosePizarro3:
> Yeah, I saw it from your implementation now. To my understanding, this runs against the SI convention.
Did you verify with Luca or anyone from Area D whether we're allowed to do this?
Don't forget that when querying via the API, the user only gets numbers, not quantities. If we change units like this, the queried data becomes meaningless...

SI convention for a Hubbard model or an electronic properties calculation? No, thanks.
I didn't discuss this with Luca, or someone from Area D. But because this does not make any sense. If you query then data, you have to either put it in joules, or know beforehand that you have to include a regularization to change units back to eV. Which queries are you referring to when becoming meaningless? Am I overlooking something?

What can't be is that 4 eV are 6.4087e-19...

JosePizarro3:
Welp, in any case, you can keep reviewing this, and we put an asterisk to the units thingy until tomorrow.

JosePizarro3:
Ok, finished here. I:

- Completely removed `HubbardModel` and `hubbard_model` to `HubbardKanamoriModel` and `hubbard_kanamori_model`.

PR #47, opened 2023-02-07
18 fhiaims extra method section

### LINKED ISSUE #18: FHIaims extra Method section
The FHIaims parser incorrectly parses the method sections and add an extra one with references plus only `electronic.method='DFT'`. You can see it in all `single_point` entries in the beta repo.

This happens at the end of `parse_configurations:parse_section`, where an if is screwing up the parsing (at least this is what I think, but further investigation is needed). @ndaelman-hu agreed to have a look.

One small point: I think this issue should be solved **after** the GW branch is merged. In that case, the lines where this happen are 1300-1305ish. Otherwise we can overwrite each others work.

ndaelman-hu:
I've removed the trailing Method section(s) in the FHIaims parser, since 
- it was giving issues with workflows (the bug part).
- it is irrelevant (there should only be one `Method` for DFT).

### DIFF
--- a/electronicparsers/fhiaims/parser.py
+++ b/electronicparsers/fhiaims/parser.py
@@ -1260,6 +1260,7 @@ class FHIAimsParser:
 
             sec_scc = sec_run.m_create(Calculation)
             sec_scc.system_ref = sec_run.system[-1]
+            sec_scc.method_ref = sec_run.method[-1]
 
             sec_energy = sec_scc.m_create(Energy)
             energy = section.get('energy', {})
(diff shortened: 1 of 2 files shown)


JosePizarro3:
Looks good to me.

Did you test it with some GeomOpt calculations to see whether everything is correctly normalized too?

ndaelman-hu:
Yes, so in the tests, it before generated 7 Methods for 6 steps, now only 1. I've attached some screenshots of the GUI version.
I got the same results with a Molecular Dynamics simulation.

@JosePizarro3 Maybe you could try this branch with a GW workflow?
Screenshot from 2023-02-08 13-43-24
Screenshot from 2023-02-08 13-42-40
Screenshot from 2023-02-08 13-42-10

JosePizarro3:
Yeah, it works. Feel free to merge when you want :+1:

ndaelman-hu:
Can you sign off?

JosePizarro3 (approved):

PR #48, opened 2023-02-09
KMesh for VASP parser

### LINKED ISSUE #36: Parse KMesh in electronic parsers
@ladinesa @ndaelman-hu we can start parsing the k points into the `KMesh` method section. In order to not step in each others feet, I created a list to keep track of which one is parsed or not.

Method name after the code name refers to the SinglePoint entries.

- [x] ABACUS
- [x] ABINIT DFT
- [x] ABINIT GW
- [x] AMS
- [ ] QuantumATK
- [ ] BigDFT
- [ ] CASTEP
- [ ] CHARMM
- [ ] CP2K
- [ ] CPMD
- [ ] CRYSTAL
- [ ] DMol3
- [ ] Elk
- [x] exciting DFT
- [x] exciting GW
- [x] exciting BSE
- [x] FHI-aims DFT
- [x] FHI-aims GW
- [ ] FLEUR
- [ ] FPLO
- [ ] GAMESS
- [ ] Gaussian
- [ ] GPAW
- [ ] Molcas
- [ ] MOPAC
- [ ] NWChem
- [x] OCEAN
- [x] Octopus
- [ ] ONETEP
- [ ] OpenMX
- [ ] ORCA
- [ ] Psi4
- [ ] Qball
- [ ] Qbox
- [x] QuantumESPRESSO
- [ ] SIESTA
- [x] soliddmft
- [ ] TURBOMOLE
- [x] VASP DFT
- [x] VASP GW
- [x] w2dynamics
- [x] Wannier90 
- [x] WIEN2k
- [ ] YAMBO

ndaelman-hu:
These commits extract the `KMesh` from the `vasprun.xml` file.
No such data is stored in the `OUTCAR`, only `KPOINTS`. However, I've added extraction of the kpoints multiplicities, which could be used to compute the `KMesh` too. I find this option preferable, since it only uses information from the `OUTCAR`.
I will have to add this normalization step after this PR though.

### DIFF
--- a/electronicparsers/vasp/parser.py
+++ b/electronicparsers/vasp/parser.py
@@ -44,7 +44,7 @@ from nomad.parsing.file_parser.text_parser import TextParser, Quantity
 from nomad.datamodel.metainfo.simulation.run import Run, Program
 from nomad.datamodel.metainfo.simulation.method import (
     Method, BasisSet, BasisSetCellDependent, DFT, HubbardKanamoriModel, AtomParameters, XCFunctional,
-    Functional, Electronic, Scf
+    Functional, Electronic, Scf, KMesh
 )
 from nomad.datamodel.metainfo.simulation.system import (
     System, Atoms
@@ -913,7 +913,7 @@ class RunContentParser(ContentParser):
             if method:
                 self._kpoints_info['x_vasp_k_points_generation_method'] = method['param']
             divisions = self._get_key_values(
-                '/modeling[0]/kpoints[0]/generation[0]/i[@name="divisions"]')
+                '/modeling[0]/kpoints[0]/generation[0]/v[@name="divisions"]')
             if divisions:
                 self._kpoints_info['divisions'] = divisions['divisions']
             volumeweight = self._get_key_values('/modeling[0]/kpoints[0]/generation[0]/i[@name="volumeweight"]')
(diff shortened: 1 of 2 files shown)


electronicparsers/vasp/parser.py
```diff
@@ -1412,6 +1422,8 @@ def parse_eigenvalues(n_calc):
                 eigs = eigs * ureg.eV
                 sec_eigenvalues = sec_scc.m_create(BandEnergies)
                 sec_eigenvalues.kpoints = kpoints
+                sec_eigenvalues.kpoints_weights = self.parser.kpoints_info.get('k_mesh_weights', [])
```
ladinesa, line 1425: i think this should be names weights and multiplicities.

ndaelman-hu: You mean internally, or the metainfo?

ndaelman-hu: If internally, I guess I should also rename `k_mesh_points`, right?

ladinesa: for the metainfo, for me it is somewhat redundant if we still prefix them with kpoints. but if luca already decided then ok.

ndaelman-hu: I'm not sure who decided on the name, still this PR is only for the parser, and more specifically the KMesh.
I just added the extraction of the `multiplicities`, since I know extract it too, and it is stored in 2 places.

I'm thinking of refactoring the normalizer, and this is when we could change the names.
Moreover, we should see how we want the k-points information structured.

ndaelman-hu: I added it to the cleaning list: https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/issues/1299
Actually, Luca and I composed a list of corrections to do for the workshop in May. I will take this one too. Can we for the time being close this PR?

ladinesa: yes ok.

PR #49, opened 2023-02-17
Ocean parser

JosePizarro3:
Initial parser for Ocean, including the creation of N entries according to the N of photons.


electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,271 @@
...
+            '[1, 0]': 'K',
+            '[2, 1]': 'L23'
+        }
+
+    def parse_system(self, archive, data):
+        sec_run = archive.run[-1]
+        sec_atoms = sec_run.m_create(System).m_create(Atoms)
+
+        if data.get('avecs'):
+            sec_atoms.lattice_vectors = data.get('avecs') * ureg.bohr
```
ladinesa, line 86: maybe some error handling necessary here in case avecs is missing

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,271 @@
...
+            sec_atoms.periodic = [data.get('avecs')[:] is not None] * 3
+            sec_atoms.lattice_vectors_reciprocal = np.array(data.get('bvecs')) / ureg.bohr
```
ladinesa, line 93: maybe error handling is necessary here in case bvecs is missing

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,271 @@
...
+
+        if data.get('znucl') and data.get('typat'):
+            sec_atoms.labels = [chemical_symbols[int(data.get('znucl')[n_at - 1])] for n_at in data.get('typat')]
+            sec_atoms.positions = data.get('xangst') * ureg.bohr
```
ladinesa, line 98: here also

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,271 @@
...
+            return keys
+        return True
+
+    def parse(self, filepath, archive, logger):
+        self.filepath = filepath
+        self.archive = archive
+        self.maindir = os.path.dirname(self.filepath)
+        self.logger = logger if logger is not None else logging
+
+        def finding_files(names):
```
ladinesa: This assumes that all files are in maindir. If this is the case, Maybe search only for absspct then simply change the file name to abslanc or photon in the loop.

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,271 @@
...
+        if data.get('avecs'):
+            sec_atoms.lattice_vectors = data.get('avecs') * ureg.bohr
+            sec_atoms.periodic = [data.get('avecs')[:] is not None] * 3
+            sec_atoms.lattice_vectors_reciprocal = np.array(data.get('bvecs')) / ureg.bohr
+
+        if data.get('znucl') and data.get('typat'):
+            sec_atoms.labels = [chemical_symbols[int(data.get('znucl')[n_at - 1])] for n_at in data.get('typat')]
+            sec_atoms.positions = data.get('xangst') * ureg.bohr
+
+    def parse_photon_polarization(self, archive, files, index):
```
ladinesa: i suggest you do the search here for photon-index

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,271 @@
...
+        archive.workflow2 = workflow
+
+    def get_mainfile_keys(self, filepath):
+        absspct_files = [f for f in os.listdir(os.path.dirname(filepath)) if f.startswith('absspct')]
+        absspct_files.sort()
+        if len(absspct_files) > 1:
+            keys = []
+            for f in absspct_files:
+                keys.append(f'BSE{f[-2:]}')
+            keys.remove('BSE01')  # removing BSE01 as this will be save in the initial call of parse()
```
ladinesa: not sure if i get this but, your mainfile is postDefaultsOceanDatafile, so essential this corresponds to the bse_workflow and the absspcts willl be the bse tasks. your bse workflow will simply connect the tasks no spectra and maybe the system

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,271 @@
...
+
+        # Program
+        sec_program = sec_run.m_create(Program)
+        sec_program.name = 'OCEAN'
+        sec_program.version = data['version'].get('.')
+        sec_program.x_ocean_commit_hash = data['version'].get('hash')
+        sec_program.x_ocean_original_dft_code = self._dft_code_map.get(data['dft'].get('program'))
+
+        # System
+        self.parse_system(archive, self.data['structure'])
```
ladinesa: error handling

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,271 @@
...
+        sec_lanczos.x_ocean_eigenvalues = data_lancz[n_dimension + 1:]
+
+    def parse_spectra_entries(self, archive, files, index):
+        # For each spectra, we parse the data in one entry
+        try:
+            data = json.load(open(self.filepath))
+        except Exception:
+            self.logger.error('Error opening json output file.')
+            data = None
+            return
```
ladinesa: what happens to the child archives then, are they also empty now? maybe put this check in get_mainfile_keys?

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,271 @@
...
+            'core_offset', 'final', 'grid']
+        for key in screen_keys:
+            setattr(sec_bse_screen, f'x_ocean_{key}', self.data['screen'].get(key))
+        for keys in screen_dicts:
+            for subkeys in self.data['screen'][keys].keys():
+                setattr(sec_bse_screen, f'x_ocean_{keys}_{subkeys}', self.data['screen'][keys].get(subkeys))
+        sec_bse_screen.x_ocean_model_flavor = self.data['screen']['model'].get('flavor')
+        # edges
+        edges = []
+        for ed in [x.split(' ') for x in self.data['calc'].get('edges')]:
```
ladinesa: error handling get('edges', [])

electronicparsers/ocean/parser.py
JosePizarro3: My initial point was not to create any BSE workflow that groups the spectra entries, but rather have N entries according to the number of N spectra files.

That's why, on the first initialization of `parse`, the archive there it correspondes to N=1 spectra, and then there are (N-1) child archives generated.

That's why I am removing here the key.

electronicparsers/ocean/parser.py
JosePizarro3: I think my idea was to define a list of possible names for files as features parsed in the parser.

Do you want me to define no attributes for this function and then loop inside the `for n in ['absspct', 'photon', 'abslanc']`?

electronicparsers/ocean/parser.py
JosePizarro3: Not sure if I understand: you mean changing the name of `index` to `photon_index` in this function?

electronicparsers/ocean/parser.py
JosePizarro3: I moved it to `parse`

electronicparsers/ocean/parser.py
JosePizarro3: done

electronicparsers/ocean/parser.py
JosePizarro3: done

electronicparsers/ocean/parser.py
JosePizarro3: done

electronicparsers/ocean/parser.py
JosePizarro3: done

JosePizarro3:
@ladinesa please test it, and let me know if you like it or something is to improve or not clear.

electronicparsers/ocean/parser.py
ladinesa: My idea is you simply loop over all the child archives and simply look for the auxiliary files corresponding to each archive.

electronicparsers/ocean/parser.py
ladinesa: This will make it harder to understand and have this unnecessary if. I suggest we make archive as photon_workflow. and the photon_archives correspond to absspct. your functions parse_spectra_entries should take in only path of abspct and from there you resolve the auxilliary files.

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,332 @@
...
+        n_dimension = int(data_lancz[0][0]) + 1
+        sec_lanczos.x_ocean_n_tridiagonal_matrix = n_dimension
+        sec_lanczos.x_ocean_scaling_factor = data_lancz[0][1]
+        matrix = [[data_lancz[1], 0.0]]
+        for n in range(2, n_dimension + 1):
+            matrix.append([data_lancz[n][0], data_lancz[n][1]])
+        sec_lanczos.x_ocean_tridiagonal_matrix = matrix
+        sec_lanczos.x_ocean_eigenvalues = data_lancz[n_dimension + 1:]
+
+    def parse_spectra_entries(self, files, index):
```
ladinesa: this should take in only path of abspsct.

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,332 @@
...
+        sec_program.x_ocean_original_dft_code = self._dft_code_map.get(self.data['dft'].get('program'))
+
+        # System
+        if not self.data.get('structure'):
+            self.logger.error('Error finding the structure in the main output file.')
+            return
+        self.parse_system(self.data.get('structure'))
+
+        # Method
+        self.parse_photon_polarization(files, index)
```
ladinesa: in parse_photon_polarization, you have something like
```
index = abspsct[-2:]
files = os.path.listdir(os.path.dirname(abspsct))
photon_files  = [f for f in files if f == f'photon-{index}'] 
```
Similarly in parse_scc you have
```
index = abspsct[-2:]
files = os.path.listdir(os.path.dirname(abspsct))
lanc_files  = [f for f in files if f.startswith('abslanc') and f.endswith(f'_{index}")] 
```

JosePizarro3:
Ok, finished with the review. Check it out if now you find it more appealing :slightly_smiling_face:

JosePizarro3:
_Note_: I had to skip the testing, as the creation of several archives breaks the pytesting. Or do you know a way around this?

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,322 @@
...
+        workflow.results.spectrum_polarization = spectra
+        photon_workflow_archive.workflow2 = workflow
+
+    def get_mainfile_keys(self, filepath):
+        absspct_files = [f for f in os.listdir(os.path.dirname(filepath)) if f.startswith('absspct')]
+        absspct_files.sort()
+        if len(absspct_files) > 0:
+            keys = []
+            for f in absspct_files:
+                keys.append(f'photon{f[-2:]}')
```
ladinesa: you mainfile key should simply be the absspct files (full path) so you do not have to do listdir again in 314

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,322 @@
...
+            data = None
+            return
+        self.data = data
+
+        # We recognize the absspct files as the main auxiliary files
+        self.aux_files = [f for f in os.listdir(self.maindir) if f.startswith('absspct')]
+        photon_archive = []
+        if self.aux_files:
+            for index in range(len(self.aux_files)):
+                self.archive = self._child_archives.get(f'photon0{index + 1}')
```
ladinesa: you should not assign self.archive here, it will confuse things most probably will cause an error, pass path to self.parse_spectra_entries

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,322 @@
...
+                keys.append(f'photon{f[-2:]}')
+            return keys
+        return True
+
+    def parse(self, filepath, archive, logger):
+        self.filepath = filepath
+        self.maindir = os.path.dirname(self.filepath)
+        self.logger = logger if logger is not None else logging
+
+        photon_workflow_archive = archive  # archive will be passed as the PhotonPolarization workflow
```
ladinesa, line 303: your

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,322 @@
...
+        n_dimension = int(data_lancz[0][0]) + 1
+        sec_lanczos.x_ocean_n_tridiagonal_matrix = n_dimension
+        sec_lanczos.x_ocean_scaling_factor = data_lancz[0][1]
+        matrix = [[data_lancz[1], 0.0]]
+        for n in range(2, n_dimension + 1):
+            matrix.append([data_lancz[n][0], data_lancz[n][1]])
+        sec_lanczos.x_ocean_tridiagonal_matrix = matrix
+        sec_lanczos.x_ocean_eigenvalues = data_lancz[n_dimension + 1:]
+
+    def parse_spectra_entries(self, index):
```
ladinesa: i think yuo do not need to pass index, instead the absscpt path so you can query the corresponding archive in child_archives. You can simply resolve the index from the path.

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,322 @@
...
+        sec_program.x_ocean_original_dft_code = self._dft_code_map.get(self.data['dft'].get('program'))
+
+        # System
+        if not self.data.get('structure'):
+            self.logger.error('Error finding the structure in the main output file.')
+            return
+        self.parse_system(self.data.get('structure'))
+
+        # Method
+        self.parse_photon_polarization(index)
```
ladinesa: pass path here

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,322 @@
...
+        self.parse_method(self.archive)
```
ladinesa: also here then resolve archive inside function

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,322 @@
...
+
+        # Calculation
+        self.parse_scc(index)
```
ladinesa: also here

electronicparsers/ocean/parser.py
```diff
@@ -0,0 +1,322 @@
...
+            sec_atoms.labels = [chemical_symbols[int(data.get('znucl')[n_at - 1])] for n_at in data.get('typat')]
+        if data.get('xangst'):
+            sec_atoms.positions = data.get('xangst') * ureg.bohr
+
+    def parse_photon_polarization(self, index):
+        sec_run = self.archive.run[-1]
+        sec_photon = sec_run.m_create(Method).m_create(Photon)
+
+        # NOT IDEAL: photonN should be in the same folder: patch due for the upload mr5PRdbVQUm-d7awz3Q9Uw
+        photon_file = [f for f in os.listdir(self.maindir) if f.startswith('photon') and f.endswith(f'{index + 1}')]
```
ladinesa: please look at the function get_exciting_files I implemented for exciting, you can copy it for now then we make it a utility function.

electronicparsers/ocean/parser.py
JosePizarro3: it is actually simplifying the other methods, that's why I kept it there. I would actually stay as it is now.

JosePizarro3: Ah, you mean it causes the error in the pytest?

ladinesa: yes since you should write your methods to self.archive(mainfile) but then you set self.archive to the child archive.

electronicparsers/ocean/parser.py
JosePizarro3: I am using parse_method for the workflow as well, that's why I am passing the archive and not the path.

ladinesa: you can access archive from child_archives by using the path as key.

electronicparsers/ocean/parser.py
JosePizarro3: done

electronicparsers/ocean/parser.py
JosePizarro3: done

electronicparsers/ocean/parser.py
JosePizarro3: done

electronicparsers/ocean/parser.py
JosePizarro3: done

electronicparsers/ocean/parser.py
JosePizarro3: done

electronicparsers/ocean/parser.py
JosePizarro3: done, changed to `self._child_archives.get(path)` with `path` being the path of the `absspct` files.

electronicparsers/ocean/parser.py
JosePizarro3: done

electronicparsers/ocean/parser.py
JosePizarro3: I think for this upload we can keep it like this, advance to the next task, and revisit it later to put `get_exciting_files` into an utils function `get_auxiliary_files`. What do you think?

JosePizarro3:
Done with the changes, please check it out. Thanks!

electronicparsers/ocean/parser.py
ladinesa: makes sense

ladinesa:
merge at will

PR #54, opened 2023-02-21
Spectra metainfo

JosePizarro3:
@ladinesa please check this, I tried to keep the same schema of OCEAN. Now the photon workflow works, which is an important first step.

We can merge this on bse-exciting, then I will work on the full DFT-Spectra workflow.

### DIFF
--- a/electronicparsers/exciting/parser.py
+++ b/electronicparsers/exciting/parser.py
@@ -1786,10 +1786,12 @@ class ExcitingParser:
 
         # Workflow
         workflow = SinglePoint2()
+        workflow.name = 'SinglePoint'
         self._child_archives.get(path).workflow2 = workflow
 
     def parse_photon_workflow(self):
         workflow = PhotonPolarization(results=PhotonPolarizationResults(), method=PhotonPolarizationMethod())
+        workflow.name = 'PhotonPolarization'
         input_structure = self.archive.run[-1].system[-1]
         input_method = self.archive.run[-1].method[-1]
         workflow.method = input_method
@@ -2000,6 +2002,7 @@ class ExcitingParser:
         sec_gw.band_structure_gw = bs_gw
 
         workflow = GW2(results=GWResults())
+        workflow.name = 'GW'
         workflow.results.dos_dft = dos_dft
         workflow.results.dos_gw = dos_gw
         workflow.results.band_structure_dft = bs_dft
@@ -2037,12 +2040,14 @@ class ExcitingParser:
 
         sec_workflow.type = 'single_point'
         workflow = SinglePoint2()
+        workflow.name = 'SinglePoint'
         sec_workflow.calculations_ref = self.archive.run[-1].calculation
         structure_optimization = self.info_parser.get('structure_optimization')
         if structure_optimization is not None:
             sec_workflow.type = 'geometry_optimization'
             sec_geometry_opt = sec_workflow.m_create(GeometryOptimization)
             workflow = GeometryOptimization2(method=GeometryOptimizationMethod())
+            workflow.name = 'GeometryOptimization'
             threshold_force = structure_optimization.get(
                 'optimization_step', [{}])[0].get('force_convergence', [0., 0.])[-1]
             sec_geometry_opt.convergence_tolerance_force_maximum = threshold_force
--- a/electronicparsers/exciting/parser.py
+++ b/electronicparsers/exciting/parser.py
@@ -34,7 +34,7 @@ from nomad.datamodel.metainfo.simulation.system import (
 )
 from nomad.datamodel.metainfo.simulation.calculation import (
     Calculation, Dos, DosValues, BandStructure, BandEnergies, Energy, EnergyEntry, Charges,
-    Forces, ForcesEntry, ScfIteration, BandGap
+    Forces, ForcesEntry, ScfIteration, BandGap, Spectra
 )
 from nomad.datamodel.metainfo.workflow import Workflow, GeometryOptimization, Task, GW as GWWorkflow
 from nomad.datamodel.metainfo.workflow2 import TaskReference, Link, Workflow as Workflow2
@@ -42,12 +42,15 @@ from nomad.datamodel.metainfo.simulation.workflow import (
     SinglePoint as SinglePoint2, GeometryOptimization as GeometryOptimization2,
     GeometryOptimizationMethod, GW as GW2, GWResults, ParticleHoleExcitations,
     ParticleHoleExcitationsMethod, ParticleHoleExcitationsResults, PhotonPolarization,
-    PhononMethod, PhononResults
+    PhotonPolarizationMethod, PhotonPolarizationResults
 )
-
-from .metainfo.exciting import x_exciting_section_MT_charge_atom, x_exciting_section_MT_moment_atom,\
-    x_exciting_section_spin, x_exciting_section_fermi_surface,\
-    x_exciting_section_atoms_group
+from .metainfo.exciting import (
+    x_exciting_section_MT_charge_atom, x_exciting_section_MT_moment_atom,
+    x_exciting_section_spin, x_exciting_section_fermi_surface,
+    x_exciting_section_atoms_group, x_exciting_exciton_calculation,
+    x_exciting_epsilon_calculation, x_exciting_sigma_calculation,
+    x_exciting_loss_calculation
+)
 
 
 re_float = r'[-+]?\d+\.\d*(?:[Ee][-+]\d+)?'
@@ -1562,7 +1565,7 @@ class ExcitingParser:
         sec_run = archive.run[0] if archive.run else archive.m_create(Run)
         return sec_run.calculation[0] if sec_run.calculation else sec_run.m_create(Calculation)
 
-    def _parse_xs_bse(self):
+    def _parse_xs_bse(self, path):
 
         def parse_exciton(data, sec_scc):
             # TODO remove n_components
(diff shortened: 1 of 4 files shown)


JosePizarro3:
Ah, and I skipped the tests, so I will work on them also in `bse-exciting` once we merge this.

electronicparsers/exciting/parser.py
```diff
@@ -2401,16 +2491,25 @@ def parse(self, filepath, archive, logger, **kwargs):
...
+            self.parse_file('input.xml', sec_method)
+            self.parse_xs(archive)
+
+            photon_archive = []
+            for child in self._child_archives:
+                if self._child_archives.get(child):
+                    self.parse_photons(child)
+                    photon_archive.append(self._child_archives.get(child))
+            # putting together all photons in the same xs_archive
+            self.parse_photon_workflow(photon_archive, archive)
```
ladinesa: you do not need to pass photon_archive since there are just the child_archives. also archive is simply self.archive

electronicparsers/exciting/parser.py
```diff
@@ -2154,8 +2246,8 @@ def parse_scf(iteration, msection):
 
         return sec_scc
 
-    def parse_system(self, section):
-        sec_run = self.archive.run[-1]
+    def parse_system(self, archive, section):
```
ladinesa: you do not need to pass archive

electronicparsers/exciting/parser.py
```diff
@@ -1691,34 +1717,105 @@ def get_data(path):
...
-        for archive in self._child_archives.values():
-            if archive.m_xpath('run[0].calculation[0].spectra[0]'):
-                spectra.append(archive.run[0].calculation[0].spectra[0])
-        sec_workflow.results.n_polarizations = len(spectra)
-        sec_workflow.results.spectrum_polarization = spectra
-        sec_workflow.outputs = [Link(name='output spectrum', section=spectrum) for spectrum in spectra]
-        self.archive.workflow2 = sec_workflow
+        outputs = []
+        for archive in photon_archive:
+            if archive.workflow2:
```
ladinesa: you do not need this if statement since you created it

electronicparsers/exciting/parser.py
```diff
@@ -1691,34 +1717,105 @@ def get_data(path):
...
+        else:
+            sec_photon = sec_run.m_create(Method).m_create(Photon)
+        sec_photon.momentum_transfer = sec_run.method[0].get('x_exciting_xs_qpointset_qpoint')
 
-        # inconsistency in the naming convention for xs input xml file
+    def parse_xs(self, archive):
+        sec_run = archive.run[-1]
         sec_method = sec_run.m_create(Method)
-
-        self.parse_file('input.xml', sec_method)
```
ladinesa, line 1700: why remove it here?

electronicparsers/exciting/parser.py
```diff
@@ -1691,34 +1717,105 @@ def get_data(path):
...
-        sec_workflow = PhotonPolarization(results=PhononResults(), method=PhononMethod())
+    def parse_photons(self, path):
+        sec_run = self._child_archives.get(path).m_create(Run)
+
+        # Program
+        sec_run.program = Program(
+            name='exciting', version=self.info_parser.get('program_version', '').strip())
+
+        # System
+        self.parse_system(self._child_archives.get(path), self.info_parser.get('groundstate'))
```
ladinesa: I think we do not need to parse system we simply put system_ref in calculation

electronicparsers/exciting/parser.py
```diff
@@ -1691,34 +1717,105 @@ def get_data(path):
...
+
+        # We parse exciting-specific quantities
+        sec_method = sec_run.m_create(Method)
+        self.parse_file('input.xml', sec_method)
```
ladinesa: also here we put method_ref

electronicparsers/exciting/parser.py
```diff
@@ -1691,34 +1717,105 @@ def get_data(path):
...
+        self.parse_xs(self._child_archives.get(path))
+
+        # Calculation
+        self.parse_spectra(path)
+
+        # Workflow
+        workflow = SinglePoint2()
+        self._child_archives.get(path).workflow2 = workflow
+
+    def parse_photon_workflow(self, photon_archive, xs_archive):
```
ladinesa: i think this can all go to parse_xs

electronicparsers/exciting/parser.py
```diff
@@ -2401,16 +2491,25 @@ def parse(self, filepath, archive, logger, **kwargs):
...
-            self.parse_xs()
-
+            self.parse_system(archive, self.info_parser.get('groundstate'))
+            sec_method = sec_run.m_create(Method)
+            self.parse_file('input.xml', sec_method)
+            self.parse_xs(archive)
+
+            photon_archive = []
+            for child in self._child_archives:
+                if self._child_archives.get(child):
```
ladinesa: no if here.

electronicparsers/exciting/parser.py
JosePizarro3: in `parse_photons` I am parsing the system for each child, hence the attribute `archive` as input.

electronicparsers/exciting/parser.py
JosePizarro3: Now, each photon polarization has two methods: first one, grouping the inputs and the photon parameters, and second one, the BSE and KMesh metainfo.

That's why I kept `parse_xs` as only parsing the BSE and KMesh, while out of it you have `parse_file(inputs)` and `parse_photon`. Then the order of parsing matters.

electronicparsers/exciting/parser.py
JosePizarro3: I will still parse system to resolve it in Results, hence in the overview.

electronicparsers/exciting/parser.py
JosePizarro3: not sure if I understand: Method[0] will contain inputs + photons. Then Method[1] will contain BSE, and also `starting_method_ref` pointing to Method[0]. Finally, calculation will have `method_ref` pointing to Method[1].

Is that what you meant?

electronicparsers/exciting/parser.py
JosePizarro3: I will actually keep `parse_xs` to parse only XS metainfo, i.e, Method XS metainfo. Also more in line with the ocean parser.

electronicparsers/exciting/parser.py
JosePizarro3: done

electronicparsers/exciting/parser.py
JosePizarro3: done

electronicparsers/exciting/parser.py
JosePizarro3: Ok, I included `method_ref` and `system_ref` in calculations, so I think we can resolve this and the previous conversation.

electronicparsers/exciting/parser.py
JosePizarro3: done

JosePizarro3:
Ok, I changed the parser according to your comments (except for the reasons I mentioned above :sweat_smile: ).

We can merge into `bse-exciting` and I will work on the full DFT-XS workflow if you are ok with that.

electronicparsers/exciting/parser.py
ladinesa: To be honest I am not sure if this actually gets resolved in the overview if we put only method_ref and system_ref. We can try and if it is not the case then we really parse them separately. My idea is that you only parse twice, once for dft method and one for xs method and simply reference them in the spectra archives.

electronicparsers/exciting/parser.py
ladinesa: ok i get it but do we really need to parse again the method for each polar

electronicparsers/exciting/parser.py
ladinesa: do we need to parse the system or simply reference it?

electronicparsers/exciting/parser.py
ladinesa: can we not simply assign the previously parsed system

electronicparsers/exciting/parser.py
ladinesa: ok fine for me now, we can parse each method and each system but we should deal with this in a separate issue

electronicparsers/exciting/parser.py
JosePizarro3: well, that makes more individual polarization more standalone... a bit redundant, but we can revisit this in a future.

ladinesa: simply copy the method from the self.archive.method

electronicparsers/exciting/parser.py
JosePizarro3: you are right, only ref

electronicparsers/exciting/parser.py
JosePizarro3: done

I also changed then parse_xs to contain also the input parameters. We simply define a photon method section to contain the photon parameters.

electronicparsers/exciting/parser.py
JosePizarro3: done

JosePizarro3:
Ok, I changed method and system to be references to the self.archive in the photon entries. Can you check it out?

electronicparsers/exciting/parser.py
```diff
@@ -1691,34 +1717,107 @@ def get_data(path):
...
-    def parse_xs(self):
-        sec_run = self.archive.run[-1]
+    def parse_polarization(self, path):
+        sec_run = self._child_archives.get(path).run[-1]
+        sec_photon = sec_run.m_create(Method).m_create(Photon)
+        # TODO check with developers if this is correct
+        sec_photon.momentum_transfer = sec_run.method[0].get('x_exciting_xs_qpointset_qpoint')
 
-        # inconsistency in the naming convention for xs input xml file
+    def parse_xs(self, archive):
```
ladinesa: you also do not need to pass archive here right, use self.archive

electronicparsers/exciting/parser.py
```diff
@@ -1691,34 +1717,107 @@ def get_data(path):
...
+
+    def parse_photons(self, path):
+        sec_run = self._child_archives.get(path).m_create(Run)
+
+        # Program
+        sec_run.program = Program(
+            name='exciting', version=self.info_parser.get('program_version', '').strip())
+
+        # System
+        sec_run.system = self.archive.run[-1].system
```
ladinesa, line 1777: can you probably do an m_copy() here? not sure if it makes a difference, ok not for the list but say, [system[0].m_copy()]

ladinesa:
satisfied now merge at will

PR #55, opened 2023-02-27
utils

JosePizarro3:
@ladinesa @ndaelman-hu what do you think about creating these utils in the electronicparsers?

I am not 100% sure whether these should be here, or in nomad directly, as they are beyond electronic parsers.

### DIFF
--- a/electronicparsers/exciting/parser.py
+++ b/electronicparsers/exciting/parser.py
@@ -20,7 +20,6 @@ import numpy as np
 import os
 import re
 import logging
-from glob import glob
 
 from nomad.units import ureg
 from nomad.parsing.file_parser import TextParser, Quantity, XMLParser, DataTextParser
--- a/electronicparsers/exciting/parser.py
+++ b/electronicparsers/exciting/parser.py
@@ -51,6 +51,9 @@ from .metainfo.exciting import (
     x_exciting_epsilon_calculation, x_exciting_sigma_calculation,
     x_exciting_loss_calculation
 )
+from ..utils import (
+    extract_section, get_files
+)
 
 
 re_float = r'[-+]?\d+\.\d*(?:[Ee][-+]\d+)?'
@@ -1134,25 +1137,6 @@ class ExcitingParser:
 
         self._xs_spectra_types = ['EPSILON', 'EXCITON', 'SIGMA', 'LOSS']
 
-    def get_exciting_files(self, pattern, filepath=None, deep=True):
-        if filepath is None:
-            filepath = self.filepath
-        # TODO do recursion until files are found
-        for _ in range(10):
-            filenames = glob(f'{os.path.dirname(filepath)}/{pattern}')
-            pattern = os.path.join('**' if deep else '..', pattern)
-            if filenames:
-                break
-
-        if len(filenames) > 1:
-            # filter files that match
-            suffix = os.path.basename(filepath).strip('INFO.OUT')
-            matches = [f for f in filenames if suffix in f]
-            filenames = matches if matches else filenames
-
-        filenames = [f for f in filenames if os.access(f, os.F_OK)]
-        return filenames
-
     def file_exists(self, filename):
         """Checks if a the given filename exists and is accessible in the same
         folder where the mainfile is stored.
@@ -1434,7 +1418,8 @@ class ExcitingParser:
         else:
             return
 
-        files = self.get_exciting_files(name, filepath=filepath if filepath is not None else self.filepath)
+        filepath = filepath if filepath is not None else self.filepath
+        files = get_files(name, filepath, 'INFO.OUT')
         if len(files) > 1:
             self.logger.warning('Found multiple files. Will read all!', data=dict(file=name))
 
@@ -1616,7 +1601,7 @@ class ExcitingParser:
 
         file_ending = path.split('EPSILON')[-1]  # Identifying files with the same ending but different type of calculation
         polarization_files = [
-            f for f in self.get_exciting_files('*BSE*.OUT', self._xs_info_file) if f.endswith(file_ending)]
+            f for f in get_files('*BSE*.OUT', self._xs_info_file, 'INFO.OUT') if f.endswith(file_ending)]
         for file in polarization_files:
             if sec_run.m_xpath('calculation'):
                 sec_scc = sec_run.calculation[-1]
@@ -1656,7 +1641,7 @@ class ExcitingParser:
 
         def get_data(path):
             # all files related to quantity at all qpoints
-            files = self.get_exciting_files(os.path.basename(path).replace('001', '*'), filepath=self._xs_info_file)
+            files = get_files(os.path.basename(path).replace('001', '*'), self._xs_info_file, 'INFO.OUT')
             data = [[], [], []]
             data_q = []
             for f in files:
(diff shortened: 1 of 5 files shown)


ladinesa:
Thanks for this. You are right they should be in utils.

electronicparsers/utils/utils.py
```diff
@@ -0,0 +1,61 @@
...
+def extract_section(source, path):
+    '''
+    Extracts the section from source given by path. Examples:
+        sec_system = extract_section(archive, 'run/system')
+        spectra = extract_section(archive, 'run/calculation/spectra')
+    '''
+    path_segments = path.split('/', 1)
+    try:
+        value = getattr(source, path_segments[0])
+        value = value[-1] if isinstance(value, list) else value
```
ladinesa: how bout traversing the list until the value is found.

JosePizarro3:
> Thanks for this. You are right they should be in utils.

I was refining it even further, by defining also some class named `BeyondDFTWorkflows` which contains all these complex GW, XS,... workflows. 

Thus, we could keep it in here for the time being.

JosePizarro3:
Ok, new one. I think this will save us a lot of time when parsing across codes, skipping all the specific-code bullshit that tends to slow us down :slightly_smiling_face:

electronicparsers/fhiaims/parser.py
```diff
@@ -1699,4 +1621,4 @@ def parse(self, filepath, archive, logger):
 
             # GW workflow
             gw_workflow_archive = self._child_archives.get('GW_workflow')
-            self.parse_gw_workflow(gw_archive, gw_workflow_archive)
+            BeyondDFTWorkflows(self.archive).parse_gw_workflow(gw_archive, gw_workflow_archive)
```
ladinesa: I think you should inherit the class, that way we have an implicit categorization of parsers.

JosePizarro3: not sure if I understand, how do you mean?

electronicparsers/utils/utils.py
JosePizarro3: Ok, check it out: what do you think?

I am not very happy, because `m_xpath` seems to be a bit slow (probably it is checking all the subsections in source). I find your previous coding more efficient.

electronicparsers/fhiaims/parser.py
ladinesa: something like ExcitingParser(BeyondDFTWorkflows). I think you then should change the name to BeyondDFTWorkflowParser

electronicparsers/utils/utils.py
ladinesa: do not use m_xpath, otherwise we do not have this helper function.

JosePizarro3: ok, changed to a try/except

electronicparsers/fhiaims/parser.py
JosePizarro3: ok, check out the changes: did I understand the suggestion correctly?

electronicparsers/fhiaims/parser.py
```diff
@@ -1699,4 +1620,4 @@ def parse(self, filepath, archive, logger):
 
             # GW workflow
             gw_workflow_archive = self._child_archives.get('GW_workflow')
-            self.parse_gw_workflow(gw_archive, gw_workflow_archive)
+            BeyondDFTWorkflowsParser(self.archive).parse_gw_workflow(gw_archive, gw_workflow_archive)
```
ladinesa: here you can simply do `self.parse_gw_workflow(gw_archive, gw_workflow_archive)`

electronicparsers/exciting/parser.py
```diff
@@ -2582,8 +2426,4 @@ def parse(self, filepath, archive, logger, **kwargs):
                 # TODO generalize to include GW step
         xs_workflow_archive = self._child_archives.get('XS_workflow')
         if xs_workflow_archive:
-            self.parse_xs_worklfow(xs_archives, xs_workflow_archive)
-
-            import json
-            with open('temp.json', 'w') as f:
-                json.dump(xs_archive.m_to_dict(), f, indent=4)
+            BeyondDFTWorkflowsParser(self.archive, self._child_archives, self._xs_spectra_types).parse_xs_workflow(xs_archives, xs_workflow_archive)
```
ladinesa: also here.

electronicparsers/exciting/parser.py
```diff
@@ -2562,7 +2406,7 @@ def parse(self, filepath, archive, logger, **kwargs):
 
             # parse gw workflow
             gw_workflow_archive = self._child_archives.get('GW_workflow')
-            self.parse_gw_workflow(gw_archive, gw_workflow_archive)
+            BeyondDFTWorkflowsParser(self.archive).parse_gw_workflow(gw_archive, gw_workflow_archive)
```
ladinesa: also here.

electronicparsers/fhiaims/parser.py
JosePizarro3: true, done

JosePizarro3:
Ok, finished with this. Could you please check it out and let me know what you think?

electronicparsers/utils/utils.py
```diff
@@ -0,0 +1,240 @@
...
+    '''
+    Extracts the section from source given by path. Examples:
+        sec_system = extract_section(archive, 'run/system')
+        spectra = extract_section(archive, 'run/calculation/spectra')
+    '''
+    section_segments = path.split('/')
+    for section in section_segments:
+        try:
+            value = getattr(source, section)
+            source = value[-1] if isinstance(value, list) else value
```
ladinesa, line 44: you should look at all entries in list to see if the intended value is there. but could also be like this, we look at only last depending on how we intend to use this function. Otherwise you could provide the whole path like run/0/system as input this is more general i think

JosePizarro3: I think we can keep it to the last section. Is it ok if I simply update the comment for `extract_section`?

ladinesa: that can also work.

ndaelman-hu: I just noticed that you are also adding a CoreHole section.
Tbh, that wasn't quite clear from the naming (I thought that it was only the addition of `utils.py`).
I will go over the metadata aspects this evening, okay?

ladinesa:
I think the important work now is to make a map of all parsers and create a heirarchy of parser base classes. With this, we can also get rid of the folder structure electronic, atomistic workflow database.

ndaelman-hu:
@ladinesa In what aspect do you mean "hierarchy"?
Should parsers start calling each other, or more something at the NOMAD side?

Tbh, I think that the central base class by parser is causing us a lot of pain, since it requires different main (and auxiliary) files to get mapped into the same structure before writing to the _archive_.
I would go rather by file parser, where each one writes to the _archive_ directly.

Choice of the file mapping should eventually lie with the user, with us only offering default settings.
Therefore, I'm not sure if we still need a parser base class. Our data structure is more complex than that of the ELNs.

ladinesa:
There is nothing really restrictive when you create baseclasses as you can inherit some functionalities or overwrite them in the parser class if you wish so. As this exercise in putting the parse_workflow_xs in utils show. If my understanding is right, your second issue is somewhat similar to what I have already created in nomad (#1363) which i am currently working on .

JosePizarro3:
Ok, I finished with this:

- I included also the parse_gw_workflow into OCEAN.
- Fixed some issues with Task / Taskold.
- And renaming a bit.

Once you are happy, approve it and we merge it :slightly_smiling_face:

ladinesa:
please merge

ndaelman-hu:
It looks good overall. We should maybe later on (after merging) see whether this belongs more in the workflow parser or normalizer.
I am happy with the idea of the `util` package though and will be extending it in the future.

PR #61, opened 2023-03-09
Bump nomad version

JosePizarro3:
@ladinesa @ndaelman-hu @fekad 

Hi, I was playing a bit trying to bump the nomad-lab version and push to Python 3.9.

As you can see, the pipeline still fails in pylint. My question for all of you would be: should we ignore E0611 and E1101 (including it in .pylintrc)? The main reason of why to do this would be that the parsers have to adapt to the tag of nomad-lab, which sometimes is behind of what we are implementing.

Risk is: we might commit mistakes when importing, but I guess this can be reflected with other errors in the pipeline?

Let me know what you think.

### DIFF
--- a/.github/workflows/python-actions.yaml
+++ b/.github/workflows/python-actions.yaml
@@ -26,7 +26,7 @@ jobs:
         python -m pylint --load-plugins=nomad.metainfo.pylint_plugin *parsers tests
     - name: mypy
       run: |
-        python -m mypy --ignore-missing-imports --follow-imports=silent --no-strict-optional *parsers tests
+        python -m mypy *parsers tests
     - name: Test with pytest
       run: |
         python -m pytest -sv tests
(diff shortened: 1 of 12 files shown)


ladinesa:
> @ladinesa @ndaelman-hu @fekad
> 
> Hi, I was playing a bit trying to bump the nomad-lab version and push to Python 3.9.
> 
> As you can see, the pipeline still fails in pylint. My question for all of you would be: should we ignore E0611 and E1101 (including it in .pylintrc)? The main reason of why to do this would be that the parsers have to adapt to the tag of nomad-lab, which sometimes is behind of what we are implementing.
> 
> Risk is: we might commit mistakes when importing, but I guess this can be reflected with other errors in the pipeline?
> 
> Let me know what you think.

There is nothing that you can do. Even if you ignore these pylint errors the tests won't go through because of the missing metainfo defs. We should really perform these tests locally before we make any commits. At the rate we are changing nomad I do not see a time when all the tests will pass finally.

JosePizarro3:
I see, you are totally right.

Do you want to keep the branch in any case? I worked on fixing some pylint and putting Python3.9 in the submodule.

PR #70, opened 2023-03-27
Wien2k parser: fix eingenvalues parsing

ondracka:
The problem is that there can be different amount of eigenvalues calculated at every k-point. However, the metainfo expects a homogeneous array, so just find the k-point with the least amount and discard the extra values at other k-points.

The current test worked just due to sheer luck. There were 116 eigenvalues at some k-points and 114 at others, but since it resulted in an exact average of 115, the test was happy. However some eigenvalues were incorrectly attributed to incorrect k-point.

Also the sorting was somehow broken, just sort the files before reading and the k-points are than guaranteed to come in order. Its way better to sort few strings instead of numpy array with potentially thousands of values anyway.

CC @ladinesa for review

### DIFF
--- a/electronicparsers/wien2k/parser.py
+++ b/electronicparsers/wien2k/parser.py
@@ -462,12 +462,16 @@ class Wien2kParser:
         nspin = self.get_nspin()
         if nspin == 1:
             files = self.get_wien2k_file(r'energy\_\d+', multiple=True)
+            # sort the files so that the k-points are read in order
+            files = sorted(files, key=lambda x: int(x.split('_')[-1]))
             if not files:
                 files = [self.get_wien2k_file('energy')]
         else:
             files = []
             for spin in ['up', 'dn']:
                 files_spin = self.get_wien2k_file(r'energy%s\_\d+' % spin, multiple=True)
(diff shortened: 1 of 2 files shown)


electronicparsers/wien2k/parser.py
```diff
@@ -484,28 +488,30 @@ def get_eigenvalues(self):
...
                             continue
                     eigenvalue = re_eigenvalue.match(line)
                     if eigenvalue:
-                        eigenvalues.append(eigenvalue.group(1))
+                        eigenvalues[-1].append(eigenvalue.group(1))
         try:
+            # There can be different amount of eigenvalues calculated at every k-point.
+            # However, the metainfo expects a homogeneous array, so just find the k-point
+            # with the least amount and discard the extra values at other k-points.
+            min_eigval_index = min(len(i) for i in eigenvalues)
```
ladinesa: please add a warning if eigenvalues are truncated.

ladinesa: please do not use i for variable name, use eig or eigenvalue

ondracka: Fixed (with basic tests as well).

PR #79, opened 2023-04-05
Improving Mesh in parsers

ndaelman-hu:

PR #80, opened 2023-04-05
Frequency and time meshes

JosePizarro3:
Implemented changes in GW, BSE, DMFT using the new `FrequencyMesh` and `TimeMesh`.

Fixed Wannier90 to populate `Method.k_mesh` and not `Projection.wannier.k_mesh`

### DIFF
--- a/electronicparsers/abinit/parser.py
+++ b/electronicparsers/abinit/parser.py
@@ -714,7 +714,7 @@ class AbinitOutParser(TextParser):
             Quantity(
                 'wigner_seitz_radius', rf'{re_n}\s*r_s\s*\=\s*{re_float}', repeats=False),
             Quantity(
-                'omega_plasma', rf'{re_n}\s*omega_plasma\s*\=\s*{re_float}\s*\[\w*',
+                'omega_plasma', rf'{re_n}\s*omega_plasma\s*\=\s*{re_float}\s*\[(?P<__unit>\w+)\]',
                 repeats=False),
         ]
 
@@ -1089,7 +1089,7 @@ class AbinitParser(BeyondDFTWorkflowsParser):
         # FrequencyMesh
         frequency_values = self.out_parser.get('screening_dataset', {}).get('frequencies', {}).get('values')
         if frequency_values is not None:
-            values = [freq[0] + freq[1] * 1j for freq in frequency_values]
+            values = [freq[0] + freq[1] * 1j for freq in frequency_values] * ureg.eV
             sec_freq_mesh = FrequencyMesh(dimensionality=1, n_points=len(frequency_values), points=values)
         else:
             freq_plasma = self.out_parser.get('gw_dataset', {}).get('omega_plasma', 0.0)
--- a/electronicparsers/abinit/parser.py
+++ b/electronicparsers/abinit/parser.py
@@ -714,7 +714,7 @@ class AbinitOutParser(TextParser):
             Quantity(
                 'wigner_seitz_radius', rf'{re_n}\s*r_s\s*\=\s*{re_float}', repeats=False),
             Quantity(
-                'omega_plasma', rf'{re_n}\s*omega_plasma\s*\=\s*{re_float}\s*\[(?P<__unit>\w+)\]',
+                'omega_plasma', rf'{re_n}\s*omega_plasma\s*\=\s*{re_float}\s*\[\w*',
                 repeats=False),
         ]
 
@@ -1089,14 +1089,13 @@ class AbinitParser(BeyondDFTWorkflowsParser):
         # FrequencyMesh
         frequency_values = self.out_parser.get('screening_dataset', {}).get('frequencies', {}).get('values')
         if frequency_values is not None:
-            freqs = frequency_values * ureg.eV
-            values = freqs[:, 0] + freqs[:, 1] * 1j
-            sec_freq_mesh = FrequencyMesh(n_points=len(freqs), values=values)
+            values = [freq[0] + freq[1] * 1j for freq in frequency_values]
+            sec_freq_mesh = FrequencyMesh(dimensionality=1, n_points=len(frequency_values), points=values)
         else:
             freq_plasma = self.out_parser.get('gw_dataset', {}).get('omega_plasma', 0.0)
             values = [0.0, freq_plasma * 1j]
-            sec_freq_mesh = FrequencyMesh(n_points=2, values=values)
-        sec_gw.m_add_sub_section(GW.frequency_mesh, sec_freq_mesh)
+            sec_freq_mesh = FrequencyMesh(dimensionality=1, n_points=2, points=values)
+        sec_method.m_add_sub_section(Method.frequency_mesh, sec_freq_mesh)
         # Screening
         occ_scr = self.out_parser.get_input_var('occ', 3, [])
         n_scr = len(occ_scr)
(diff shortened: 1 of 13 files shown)


JosePizarro3:
I think I don't need review for this, as it is a refactoring, but @ladinesa can you check everything is good? Thanks!

JosePizarro3:
@ndaelman-hu this interests you.

electronicparsers/abinit/parser.py
```diff
@@ -1089,14 +1089,13 @@ def parse_gw(self):
...
-            values = freqs[:, 0] + freqs[:, 1] * 1j
-            sec_freq_mesh = FrequencyMesh(n_points=len(freqs), values=values)
+            values = [freq[0] + freq[1] * 1j for freq in frequency_values] * ureg.eV
+            sec_freq_mesh = FrequencyMesh(dimensionality=1, n_points=len(frequency_values), points=values)
         else:
             freq_plasma = self.out_parser.get('gw_dataset', {}).get('omega_plasma', 0.0)
             values = [0.0, freq_plasma * 1j]
-            sec_freq_mesh = FrequencyMesh(n_points=2, values=values)
-        sec_gw.m_add_sub_section(GW.frequency_mesh, sec_freq_mesh)
+            sec_freq_mesh = FrequencyMesh(dimensionality=1, n_points=2, points=values)
```
ladinesa, line 1097: Should dimensionality be a derived quantity? This should be a comment in gitlab.

JosePizarro3: It is always 1. Overwritten by the normalizer anyways, just set it here to keep consistency with the dimensions of `points`.

ladinesa:
Everything looks good.

electronicparsers/abinit/parser.py
```diff
@@ -1089,14 +1089,13 @@ def parse_gw(self):
         # FrequencyMesh
         frequency_values = self.out_parser.get('screening_dataset', {}).get('frequencies', {}).get('values')
         if frequency_values is not None:
-            freqs = frequency_values * ureg.eV
-            values = freqs[:, 0] + freqs[:, 1] * 1j
-            sec_freq_mesh = FrequencyMesh(n_points=len(freqs), values=values)
+            values = [freq[0] + freq[1] * 1j for freq in frequency_values] * ureg.eV
+            sec_freq_mesh = FrequencyMesh(dimensionality=1, n_points=len(frequency_values), points=values)
```
ndaelman-hu, line 1093: You can leave `dimensionality` out (unless it deviates from 1), since it is handled by the normalizer

ndaelman-hu: We can do the same with `n_points`, btw.
I would only set `n_points` if there's really a constraint from some other source.
This here is a good example, since it comes from `frequency_values` rather than `values`

electronicparsers/fhiaims/parser.py
```diff
@@ -667,7 +667,7 @@ def str_to_md_control_in(val_in):
                 'k_grid',
                 rf'{re_n} *k\_grid\s*([\d ]+)', repeats=False),
             Quantity(
-                'freq_grid_type', rf'{re_n}\s*Initialising transformed\s([\w\-]+)\s*time and frequency grids',
+                'freq_grid_type', rf'{re_n}\s*Initialising\s*([\w\-\s]+)\s*time and frequency grids',
```
ndaelman-hu: I think that you can simplify `\s*([\w\-\s]+)\s*` to `([\w\-\s]+)`: space is now incorporated in the matching group.
At most, you have to apply `strip` as a string operation.

electronicparsers/w2dynamics/parser.py
```diff
@@ -210,6 +211,16 @@ def parse_method(self, data):
         sec_dmft.n_correlated_orbitals = corr_orbs_per_atoms
         sec_dmft.n_correlated_electrons = occ_per_atoms
         sec_dmft.impurity_solver = 'CT-HYB'
+        # FrequencyMesh
+        if sec_run.m_xpath('x_w2dynamics_axes.x_w2dynamics_iw') is not None:
+            iw = sec_run.x_w2dynamics_axes.x_w2dynamics_iw * 1j * ureg.eV
```
ndaelman-hu, line 216: So, a general comment, where I would also like the opinion of @ladinesa :
When mapping parser/application-specific quantities into general NOMAD quantities, I remove the former from the parser metainfo.

The reason is that, to my understanding, these parser/application-specific quantities will still be printed to the archive, clogging it up with double data. What do you think?

ladinesa: we should try not to define parser-specific quantities if we can map then to the nomad defs.

ndaelman-hu:
> `sec_k_mesh.grid = self.out_parser.get('k_grid', [1, 1, 1])`

So, in the FHIaims parser, you default the `k_mesh.grid` to the origin.
To my understanding, AIMS does not have a guaranteed PBC (check the `atom_frac` tag in the manual).
Actually, this would be a case with variable dimensionality, as is CP2K.
From how I understood @ladinesa , we should for example write `(1, 1, 0)` in the 2D case. Anyhow, the default seems too restrictive.

ndaelman-hu:
@JosePizarro3 Okay, done with the review. Minor comments, overall, please resolve the threads and we can merge.

electronicparsers/w2dynamics/parser.py
ndaelman-hu: Could we see to add these guidelines somewhere in the parser docs?

JosePizarro3: In this case it was a mere copy-paste being lazy. In the mainfile hdf5, there is a group called ".axes" which is mapped directly into those code-specific metainfo in `Run`, and after that, mapped into the nomad defs. This happens for other quantities.

But yeah, I agree with you both.

I can change this in a separate issue, as I should revisit w2dynamics and soliddmft to do so.

JosePizarro3:
> > `sec_k_mesh.grid = self.out_parser.get('k_grid', [1, 1, 1])`
> 
> So, in the FHIaims parser, you default the `k_mesh.grid` to the origin. To my understanding, AIMS does not have a guaranteed PBC (check the `atom_frac` tag in the manual). Actually, this would be a case with variable dimensionality, as is CP2K. From how I understood @ladinesa , we should for example write `(1, 1, 0)` in the 2D case. Anyhow, the default seems too restrictive.

I can leave it empty to be `None`

ndaelman-hu:
@JosePizarro3 Do we merge?

JosePizarro3:
> @JosePizarro3 Do we merge?

I think not. I should do the minor changes. Furthermore, we should try to coincide with #79 and then update the ref in gitlab right? Otherwise, the days diff will cause a lot of troubles.

electronicparsers/fhiaims/parser.py
JosePizarro3: Ok, done, I also changed the KMesh section in GW.

ndaelman-hu:
Please feel free to add the changes that you feel need adding.
Regarding merging inot #79 , it's not that big of a deal: I can do the merge cleanly in VS Code, so you don't have to worry about that.

ndaelman-hu:
Can you change the target branch, then I apply the merge.

JosePizarro3:
Ok, done. Please, double-check it before merging.

PR #83, opened 2023-04-14
Update bandgap

ndaelman-hu:
Update band gap extraction in line with https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1185
Actually, we should decide on whether we also want `band_gaps` from the parsers.


ndaelman-hu:
@JosePizarro3 So, if we decide to also capture band gaps from parsers, we can start with _exciting_.
I've set up the right section, but I'll leave it up to you how you want the provenance:

1. The best option would be to add special sections in `ElectronicStructureProvenance`, but I'd need to know the suggested names and to which section they'd point.
2. We use the label string, suggested by Pepe.

electronicparsers/abinit/parser.py
```diff
@@ -1392,9 +1392,8 @@ def parse_meshes(section, keys, value):
             selfenergy_data = np.array(selfenergy.get('data', []))
             if selfenergy.get('params'):
                 sec_k_eigen.n_spin_channels = selfenergy.get('params')[0][-1]
-                sec_gap = sec_k_eigen.m_create(BandGap)
+                sec_gap = sec_k_eigen.m_create(BandGapDepricated)
                 sec_gap.value = selfenergy.get('params')[1][-1] * ureg.eV  # KS gap
-                sec_gap.value_fundamental = selfenergy.get('params')[2][-1] * ureg.eV  # QP gap
```
JosePizarro3, line 1397: we should store the QP gap, i.e., this line should be kept, the one before deleted.

electronicparsers/exciting/parser.py
```diff
@@ -1957,16 +1957,9 @@ def parse_gw(self):
...
-            fundamental_band_gap = self.info_gw_parser.get('fundamental_band_gap', None)
-        sec_gap = sec_scc.eigenvalues[-1].m_create(BandGap)
-        if fundamental_band_gap is not None:
-            sec_gap.value_fundamental = fundamental_band_gap
-
-        optical_band_gap = self.info_gw_parser.get('optical_band_gap', None)
-        if optical_band_gap is not None:
-            sec_gap.value_optical = optical_band_gap
+        # example band gap setup
+        sec_gap = sec_scc.m_create(BandGap)
```
JosePizarro3: Mainly for me (you can leave it out if you don't want to include it now):
```
gw_band_gap = self.info_gw_parser.get('direct_band_gap')
if gw_band_gap is None:
    gw_band_gap = self.info_gw_parser.get('fundamental_band_gap')
if gw_band_gap:
    sec_gap = sec_scc.m_create(BandGap)
    sec_gap.value = gw_band_gap * ureg.eV  
```

electronicparsers/exciting/parser.py
```diff
@@ -1957,16 +1957,9 @@ def parse_gw(self):
...
+        sec_gap.provenance = ElectronicStructureProvenance()
```
JosePizarro3: This is then enough to track whether it comes from BS or DOS? I am not sure if I understand it... as for now and where it is, it is coming from neither of these.

Most probably I am misunderstanding something.

tests/test_excitingparser.py
```diff
@@ -176,8 +176,6 @@ def test_gw(silicon_gw):
 
     # Check GW properties
     assert approx(sec_sccs[0].energy.fermi.magnitude, 1.09865567e-19)
-    assert approx(sec_sccs[0].eigenvalues[0].band_gap[0].value_fundamental.magnitude, 3.42913865e-19)
```
JosePizarro3, line 179: Add this back, with `value` instead of `value_fundamental` and in the corresponding `band_gap` section inside `sec_sccs[0]`

JosePizarro3 (commented):
I guess you are still working on this. But here you have some initial comments. Thanks!

electronicparsers/abinit/parser.py
ndaelman-hu: Okay, 2 questions:

1. Is the QP guaranteed to exist? Else, I store it preferentially, using the KS only as a backup.
2. Should we use the `provenance` to clarify whether it is QP or KS (similar to the point raised above)? I remember you mentioning that we no longer need the distinction.

electronicparsers/exciting/parser.py
ndaelman-hu: No, it isn't enough yet. What I fill in is contingent on the response to my question above: **do we go with string or section for the provenance here?** The latter would be cleaner, I guess, but I'm unsure if there's anything we could point to. Alternatively, we could also omit direct extraction from the parsers.

electronicparsers/abinit/parser.py
JosePizarro3: 1. Yes, QP == GW gap. I mean, it is the calculated from GW. KS is the DFT one.
2. Don't think so... but I am confused: this kind of info is contained in the method. For the DFT entry, you will have the KS gap, for the GW entry the QP gap.

Here both were stored prior to your changes as it is a common practice to keep both in outputs. I should check the other codes to see.

tests/test_excitingparser.py
JosePizarro3: I think you should delete `eigenvalues[0]` as well as populate that band_gap. I can take care of this part, no problem.

electronicparsers/exciting/parser.py
JosePizarro3: Should be the section method.gw or dft or bse... that's what is confusing me 😄 

But what about: at the parser level we always point to the corresponding method section. At the normalizer, if not band gap is found, then it checkes dos and band structure and then it points to the corresponding.

ndaelman-hu: > Should be the section method.gw or dft or bse... that's what is confusing me 

I was more so thinking of just labelling it as `'parser'`, to distinguish it from `dos` or `band_structure`.
I get the impression that your suggestion is trying to impart some added semantics?
Are we running up against this distinction between optical and fundamental band gap again?

**It's probably best that I constrain `label` to some enum to avoid abuse.
There's no set list, so feel free to start a discussion on what should be included over at the Gitlab, along with Pepe.**

ndaelman-hu: Regarding the parsers, I was just updating those that had some form band gap extraction.
This would also serve as an example. However, if it takes too long, I simply drop support.

tests/test_excitingparser.py
ndaelman-hu: Applied

electronicparsers/exciting/parser.py
JosePizarro3: Can you add these lines? Or do you want me to do it? I still don't understand what I have to write in `sec_gap.provenance`

tests/test_excitingparser.py
JosePizarro3: Keep it open please, I'd like this to be included when we merge, and not separate it from this pr.

ndaelman-hu: I'm confused now. Could you please be more specific with "it"?
Do you mean the unresolving the conversation, restoring the error, or something else?

electronicparsers/exciting/parser.py
ndaelman-hu: They have been added, that's why I resolved the conversation.

I just had to modify them slightly, since they were generating errors:
`gw_band_gap * ureg.eV` multiplies a `Quantity` which already has a unit with another unit.

The only other change is to use the 2-parameter version of `get()`, where the 2nd sets a default.
As such, I contract the 3-liner into a 1-liner, which only split for PEP styling.

electronicparsers/abinit/parser.py
ndaelman-hu: I can **store both**, but then we need some way (`provenance`) to **distinguish which is which**.
Tbh, the original scope of `provenance` was constrained to the normalizers.
Experiments surely have to be included, and I think it's fine to include parsers too.
The question is whether we want even more precision than the `parser` label.

**If, however, one band gap only goes to DFT and the other only to GW,** then the entry (method) alone suffices.
In that's case, I'd run with your initial suggestion on swapping the removal.

JosePizarro3: Yeah, that's my point: Method will anyways establish the provenance. Ok then, the value from KS is irrelevant in this entry, and one can get rid of it.

electronicparsers/exciting/parser.py
JosePizarro3: yeah, nice, thanks!

tests/test_excitingparser.py
JosePizarro3: ah, lol, when I wrote the comment, the changes were not appearing. I see now you included a test for the band_gap[0].value so I am ok with that 🙂

electronicparsers/abinit/parser.py
ndaelman-hu: Done.

I went with `parser` for provenance. Provenance here should be understood, how we got to the band gap (i.e. derived from a DOS, band structure, user input, etc.), not the method used to compute it.
We'll codify it into a Enum in a separate issue, where I'll also extend the description.
I'll apply this everywhere else too now.

ndaelman-hu:
Do you think that we need some `band_gap` tests?

electronicparsers/abinit/parser.py
ndaelman-hu: One more question here: you're writing out the GW band gap to the band structure.
In the update scheme, the approach would now be to write it directly to Calculation.
Actually, the section under BandEnergies does not have a provenance and will become depricated soon.

I suggest to migrate it then. Now, I know that GW is a bit more involved. I can tackle it in this PR with your guidance.
Else, you'll have to do it yourself some other time.

ndaelman-hu: Actually, this is similar to the approach in the exciting parser

JosePizarro3: > I went with parser for provenance.

Perfect for me.

> One more question here: you're writing out the GW band gap to the band structure.

It is because the value of this band_gap is per kpoint, hence it is storing the energy differences. I am not 100% sure this should be resolved as the BandGap at the end of the day to show in the overview

JosePizarro3: > Actually, this is similar to the approach in the exciting parser

Sadly not, exciting stores the full band gap, ABINIT in each kpoint if present

electronicparsers/exciting/parser.py
```diff
@@ -1957,16 +1957,13 @@ def parse_gw(self):
...
-        optical_band_gap = self.info_gw_parser.get('optical_band_gap', None)
-        if optical_band_gap is not None:
-            sec_gap.value_optical = optical_band_gap
+        # example band gap setup
+        gw_band_gap = self.info_gw_parser.get('direct_band_gap',
+                        self.info_gw_parser.get('fundamental_band_gap'))
+        if gw_band_gap is not None:
+            sec_gap = sec_scc.m_create(BandGap)
+            sec_gap.value = gw_band_gap
+            sec_gap.provenance = ElectronicStructureProvenance()
```
JosePizarro3, line 1966: here should be then?
`sec_gap.provenance = ElectronicStructureProvenance(label='parser')`

JosePizarro3: here should be?
`sec_gap.provenance = ElectronicStructureProvenance(label='parser')`

electronicparsers/abinit/parser.py
ndaelman-hu: > It is because the value of this band_gap is per kpoint, hence it is storing the energy differences. I am not 100% sure this should be resolved as the BandGap at the end of the day to show in the overview

Every day you learn something new: it never crossed my mind that there would be codes writing it out like that. We should see to add better links to the relevant documentation, but some other time.
In that case, I would remove it fully now: as long as the band structure is extracted, the normalizer will assign a band gap.

electronicparsers/abinit/parser.py
```diff
@@ -1392,9 +1393,9 @@ def parse_meshes(section, keys, value):
             selfenergy_data = np.array(selfenergy.get('data', []))
             if selfenergy.get('params'):
                 sec_k_eigen.n_spin_channels = selfenergy.get('params')[0][-1]
-                sec_gap = sec_k_eigen.m_create(BandGap)
-                sec_gap.value = selfenergy.get('params')[1][-1] * ureg.eV  # KS gap
-                sec_gap.value_fundamental = selfenergy.get('params')[2][-1] * ureg.eV  # QP gap
+                sec_gap = sec_k_eigen.m_create(BandGapDeprecated)
```
JosePizarro3: is this working? I was expecting it to m_create now `x_abinit_qp_band_gaps_params` instead of `BandGapDeprecated`, right?

JosePizarro3 (approved):
Minor doubt / check, then you can merge

PR #84, opened 2023-04-17
Fix system_ref in qe parser

ladinesa:


JosePizarro3 (approved):
Please, drop a comment to https://github.com/nomad-coe/nomad/issues/65, and say that we are optimizing the parsing of w2dynamics, as the parsed result is too big for the archive. He should see that, when this is in the beta, the parsing might still fail for w2dynamics (Tho I am not 100% sure of it)

PR #89, opened 2023-04-24
BSE changes

JosePizarro3:
Adding `BSE.type` and `BSE.solver` to exciting and OCEAN, as well as small fix on solid_dmft parser.


JosePizarro3:
Minor changes; but I need to wait for the review in nomad, so this might be sitting for a bit here.

PR #91, opened 2023-04-25
Fix scc w2dynamics

JosePizarro3:
When optimizing w2dynamics, I introduced an error from allowing `farg` to be `'wb'`. Here I defined it by default to `'r+b'` as it is always based on the w2dynamics hdf5 mainfile (so we just need to provide paths to the hdf5 groups and attributes, not rewrite the file itself).

@ladinesa does this make sense for you? Please, approve it if so.

### DIFF
--- a/electronicparsers/w2dynamics/parser.py
+++ b/electronicparsers/w2dynamics/parser.py
@@ -69,6 +69,7 @@ class W2DynamicsParser:
         self._calculation_type = 'dmft'
         self.wout_parser = WOutParser()
         self.hr_parser = HrParser()
+        self.data = None
 
         self._hubbard_kanamori_map = {
             'u': 'u',
@@ -303,12 +304,12 @@ class W2DynamicsParser:
                 n_ineq += 1
         n_atoms = sec_run.method[-1].dmft.n_atoms_per_unit_cell
 
+        filename = os.path.join(os.path.dirname(self.filepath.split("/raw/")[-1]), self.mainfile)
+        farg = 'r+b'  # Always reading the hdf5 mainfile
         for key in calc_keys:
             if key not in self.data.keys():
                 continue
             sec_scf_iteration = sec_scc.m_create(ScfIteration)
-            filename = os.path.join(os.path.dirname(self.filepath.split("/raw/")[-1]), self.mainfile)
-            farg = 'r+b' if os.path.isfile(os.path.join(os.path.dirname(self.filepath), filename)) else 'wb'
             if self.archive.m_context:
                 with self.archive.m_context.raw_file(filename, farg) as f:
                     for subkey in self.data.get(key).keys():


JosePizarro3:
Btw, after talking with @TLCFEM there is another way of getting the path to the upload, meaning one can use:
```
from nomad.files import UploadFiles
...
upload_id = archive.metadata.upload_id
upload_path = UploadFiles.get(upload_id)
```

What do you think is better? `m_context.raw_file`, `UploadFiles` or both are equivalent?

TLCFEM:
Of course the later is less safer. If it gives more it should be fine to use it with great care.

ladinesa:
> Btw, after talking with @TLCFEM there is another way of getting the path to the upload, meaning one can use:
> 
> ```
> from nomad.files import UploadFiles
> ...
> upload_id = archive.metadata.upload_id
> upload_path = UploadFiles.get(upload_id)
> ```
> 
> What do you think is better? `m_context.raw_file`, `UploadFiles` or both are equivalent?

I am not quite sure if the metadata is already filled up at this point. so imo it is still safer to do the former.

TLCFEM:
> I am not quite sure if the metadata is already filled up at this point. so imo it is still safer to do the former.

Basic info such as id has been populated there before normalizing, after that, additional info like processing logs/errors will be populated again.

ladinesa:
> > I am not quite sure if the metadata is already filled up at this point. so imo it is still safer to do the former.
> 
> Basic info such as id has been populated there before normalizing, after that, additional info like processing logs/errors will be populated again.

Great. thanks for the info. But @JosePizarro3 be careful when using it, I still think that this should not be exposed at the parser level,

TLCFEM:
Strongly agree with this for safety consideration.

JosePizarro3:
Thanks both for the info. @ladinesa can I merge?

> But @JosePizarro3 be careful when using it, I still think that this should not be exposed at the parser level.

I think it will depend on cases. How do you do it when handling workflows such as in phonopy? Ref to the file with `resolve_archive`, right? But there are cases where in workflows you clearly have only 1 mainfile at level=0, while several at level=1 for which `UploadFiles` can be the way of going if we know a priory the parser for the level=0.

I can show some specific situation where this would happen. Everything with `try`, of course.

ladinesa:
yes please merge, we can discuss this in a separate issue.

electronicparsers/w2dynamics/parser.py
```diff
@@ -303,12 +304,12 @@ def parse_scc(self):
                 n_ineq += 1
         n_atoms = sec_run.method[-1].dmft.n_atoms_per_unit_cell
 
+        filename = os.path.join(os.path.dirname(self.filepath.split("/raw/")[-1]), self.mainfile)
+        farg = 'r+b'  # Always reading the hdf5 mainfile
```
ladinesa, line 308: maybe just put this directly in 314

JosePizarro3: everytime is defining this within the for loop. Is it not better here?

ladinesa: ah i see i did not see it is also used somewhere else. ok then everythings fine

PR #95, opened 2023-05-08
Update basis set

ndaelman-hu:
Applies the updated `BasisSet` metainfo introduced in https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1264.
Heavily extends the LAPW basis sets in _exciting_, fleur, wien2k (but not elk).

### DIFF
--- a/electronicparsers/abacus/metainfo/abacus.py
+++ b/electronicparsers/abacus/metainfo/abacus.py
@@ -161,12 +161,12 @@ class x_abacus_section_specie_basis_set(MSection):
         ''')
 
 
-class x_abacus_section_basis_sets(MSection):
+class BasisSet(simulation.method.BasisSet):
     '''
     section for numerical atomic orbitals of ABACUS
     '''
 
-    m_def = Section(validate=False)
+    m_def = Section(validate=False, extends_base_section=True)
 
     x_abacus_basis_sets_delta_k = Quantity(
         type=np.dtype(np.float64),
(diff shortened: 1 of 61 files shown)


JosePizarro3:
> Everything should be updated before merging. Electronic parsers are already done, normalization mostly so. I'll trace the errors and tackle this.

`utils/utils.py` should be also updated in line 129 (when extracting the `run/method/basis_set` for the `workflow2.method`), as well as its corresponding metainfo in workflow. I wonder whether it should be moved then to the `WorkflowNormalizer` anyways, similar to the dos, band_structure, and band_gap. Let me know.

ndaelman-hu:
Extended the change to `GWMethod`. Thank you for brining it up, though pls note that even without these corrections, it passes the tests. I recommend that you consider adding more tests in this regime, diminishing the burden on the original author to follow up.

Regarding the placement, I quite neutral here, though your suggestion sounds good. Since this is more of a lay-out consideration, I consider it outside the scope of this PR. I suggest looking into it after @ladinesa has consolidated his workflow changes.

ndaelman-hu:
Additional note: `basis_set` and `electrons_representation` are both repeated sections in `run.method`, but not in `gw_method`. Should I change this?

JosePizarro3:
> though pls note that even without these corrections, it passes the tests. I recommend that you consider adding more tests in this regime, diminishing the burden on the original author to follow up.

Yes, it is because the GW workflow archive it is not passed in the tests, but rather the main first one, i.e., the DFT archive. Maybe @ladinesa can tell us if it is possible to test specific child archives in the testing of each parser? But I remember we cannot.

> Additional note: basis_set and electrons_representation are both repeated sections in run.method, but not in gw_method. Should I change this?

Why this section is repeated? For me it is ok to keep it the same, so you can change that if you don't mind. Thanks!

PR #104, opened 2023-05-16
Auto DMFT workflows

JosePizarro3:
Hi @ladinesa this is the changes attempting to recognize the DMFT workflow for w2dynamics from a Wannier90 entry.

The main idea is that, if using the upload_id together with the nearest scanned Wannier90 mainfile, this will be recognize anyway in the w2dynamics as the Wannier90 original calculation.

Let me know what do you think, whether we should not use `UploadFiles` and `Entry`, and if so, alternatives to this. Anything we decide should be somehow a standard when trying to resolve automatically these complex workflows 🙂


ladinesa:
I suggest we use nomad.search, something like 
```
search(
            owner='visible',
            user_id=archive.metadata.main_author.user_id,
            query={'upload_id': archive.metadata.upload_id},
            required=MetadataRequired(include=['entry_id'])
        ).data

**Comment by JosePizarro3** [comment]
Still I need to resolve the upload_path, for which I need `UploadFiles`.

**Comment by JosePizarro3** [comment]
Furthermore, I am not sure why `search` is better than `Entry.objects`. Why you are suggesting this?

**Comment by ladinesa** [comment]
> Furthermore, I am not sure why `search` is better than `Entry.objects`. Why you are suggesting this?

Because this is the interface to this core functionalities, which I think should not be exposed. You can add any metadata quantities to MetadataRequired such as mainfile

**Comment by JosePizarro3** [comment]
Still I need the upload_path to put it in `mainfile = f'{upload_path}/raw/{entry_archive.metadata.mainfile}'` as I am using `get_files` to get the abspath

**Comment by ladinesa** [comment]
> Still I need the upload_path to put it in `mainfile = f'{upload_path}/raw/{entry_archive.metadata.mainfile}'` as I am using `get_files` to get the abspath

Is it not possible to do path.split('raw')[-1] and compare only this?

**Comment by JosePizarro3** [comment]
> Is it not possible to do path.split('raw')[-1] and compare only this?

Ok, it has to be with `'raw/'` but it worked. Thanks 👍🏻

File: electronicparsers/w2dynamics/parser.py [hunk 1]
```diff
@@ -435,3 +449,28 @@ def parse(self, filepath, archive, logger):
...
+            wannier90_path = wannier90_files[-1].split('raw/')[-1]
+            try:
+                upload_id = self.archive.metadata.upload_id
+                search_ids = search(
+                    owner='visible',
+                    user_id=self.archive.metadata.main_author.user_id,
+                    query={'upload_id': upload_id},
+                    required=MetadataRequired(include=['entry_id', 'mainfile'])
+                ).data
+                metadata = [[sid['entry_id'], sid['mainfile']] for sid in search_ids]
```
ladinesa, line 464: could simply be for 
```py
for entry_id, mainfile in search_ids:

Comment (JosePizarro3) [reply electronicparsers/w2dynamics/parser.py:464]: Nope, not working.

Comment (ladinesa) [reply electronicparsers/w2dynamics/parser.py:464]: Ah yes, it is a dict, so could be for 
```py
for search_id in search_ids:
    seach_id['entry_id']
   ....
```
to avoid the ifs and additional loop

electronicparsers/w2dynamics/parser.py
```diff
@@ -158,7 +162,7 @@ def parse_system(self):
             if structure.get('positions') is not None:
                 sec_atoms.positions = structure.get('positions') * ureg.angstrom
         else:  # TODO parse specific lattice model: discuss it Jonas Schwab
-            self.logger.warning('Wannier90 output files not found in the same folder.')
+            self.logger.warning('Wannier90 output files not found in the same folder: I cannot resolve the system metainfo.')
```
ladinesa: folder, cannot resolve system metainfo

electronicparsers/w2dynamics/parser.py
```diff
@@ -184,7 +188,7 @@ def parse_input_model(self, data):
             sec_hopping_matrix.value = np.reshape(
                 full_hoppings, (sec_hopping_matrix.n_wigner_seitz_points, sec_hopping_matrix.n_orbitals * sec_hopping_matrix.n_orbitals, 7))
         else:  # TODO parse specific lattice model
-            pass
+            self.logger.warning('Wannier90 _hr.dat files not found in the same folder: I cannot resolve the initial model metainfo.')
```
ladinesa: folder, cannot resolve initial model metainfo.

electronicparsers/w2dynamics/parser.py
JosePizarro3: done

electronicparsers/w2dynamics/parser.py
JosePizarro3: done

JosePizarro3:
I merge then, thanks for the review.

PR #107, opened 2023-05-18
Add native tiers

ndaelman-hu:
Accompanies MR https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1283

### DIFF
--- a/electronicparsers/castep/parser.py
+++ b/electronicparsers/castep/parser.py
@@ -626,7 +626,7 @@ class CastepParser:
             'Dispersion corrected final free energy* (Ecor-TS)': 'x_castep_total_dispersion_corrected_free_energy',
             'NB dispersion corrected est. 0K energy* (Ecor-0.5TS)': 'x_castep_disp_corrected_energy_total_T0',
             'basis set parameters': {
-                'basis set accuracy': 'native_tier',  # default: FINE
+                'basis set accuracy': 'native_tier',  # https://www.tcm.phy.cam.ac.uk/castep/documentation/WebHelp/content/modules/castep/keywords/k_basis_precision_castep.htm
                 'size of standard grid': 'x_castep_size_std_grid',
                 'size of   fine   gmax': 'x_castep_size_fine_grid'},
             'phonon parameters': {
@@ -758,6 +758,7 @@ class CastepParser:
         basis_parameters = self.out_parser.get('title', {}).get('basis set parameters', {})
         sec_method.electrons_representation.append(
             BasisSetContainer(
+                native_tier=basis_parameters.get('basis set accuracy', 'FINE'),
                 type='plane waves',
                 scope=['wavefunction'],
                 basis_set=[
--- a/electronicparsers/castep/parser.py
+++ b/electronicparsers/castep/parser.py
@@ -626,6 +626,7 @@ class CastepParser:
             'Dispersion corrected final free energy* (Ecor-TS)': 'x_castep_total_dispersion_corrected_free_energy',
             'NB dispersion corrected est. 0K energy* (Ecor-0.5TS)': 'x_castep_disp_corrected_energy_total_T0',
             'basis set parameters': {
+                'basis set accuracy': 'native_tier',  # default: FINE
                 'size of standard grid': 'x_castep_size_std_grid',
                 'size of   fine   gmax': 'x_castep_size_fine_grid'},
             'phonon parameters': {
--- a/electronicparsers/orca/metainfo/orca.py
+++ b/electronicparsers/orca/metainfo/orca.py
@@ -238,13 +238,6 @@ class Method(simulation.method.Method):
         -
         ''')
 
-    x_orca_energy_change_tolerance = Quantity(
-        type=np.dtype(np.float64),
-        shape=[],
-        description='''
-        -
-        ''')
-
     x_orca_exchange_functional = Quantity(
         type=str,
         shape=[],
@@ -578,6 +571,17 @@ class Method(simulation.method.Method):
         sub_section=SectionProxy('x_orca_section_functionals'),
         repeats=True)
 
+class Scf(simulation.method.Scf):
+
+    m_def = Section(validate=False, extends_base_section=True)
+
+    x_orca_last_max_density_change = Quantity(
+        type=np.dtype(np.float64),
+        shape=[],
+        description='''
+        -
+        ''')
+
 
 class ScfIteration(simulation.calculation.ScfIteration):
 
@@ -639,13 +643,6 @@ class ScfIteration(simulation.calculation.ScfIteration):
         -
         ''')
 
-    x_orca_last_energy_change_tolerance = Quantity(
-        type=np.dtype(np.float64),
-        shape=[],
-        description='''
-        -
-        ''')
-
     x_orca_last_energy_change = Quantity(
         type=np.dtype(np.float64),
         shape=[],
@@ -653,28 +650,7 @@ class ScfIteration(simulation.calculation.ScfIteration):
         -
         ''')
 
-    x_orca_last_max_density_change = Quantity(
-        type=np.dtype(np.float64),
-        shape=[],
-        description='''
-        -
-        ''')
-
-    x_orca_last_max_density_tolerance = Quantity(
-        type=np.dtype(np.float64),
-        shape=[],
-        description='''
-        -
-        ''')
-
     x_orca_last_rms_density_change = Quantity(
-        type=np.dtype(np.float64),
-        shape=[],
-        description='''
-        -
-        ''')
-
-    x_orca_last_rms_density_tolerance = Quantity(
         type=np.dtype(np.float64),
         shape=[],
         description='''
--- a/electronicparsers/orca/metainfo/orca.py
+++ b/electronicparsers/orca/metainfo/orca.py
@@ -40,7 +40,6 @@ class x_orca_atom_positions(MSection):
         type=str,
         shape=[],
         description='''
-        -
         ''')
 
     x_orca_atom_positions_x = Quantity(
@@ -48,7 +47,6 @@ class x_orca_atom_positions(MSection):
         shape=[],
         unit='meter',
         description='''
-        -
         ''')
 
     x_orca_atom_positions_y = Quantity(
@@ -56,7 +54,6 @@ class x_orca_atom_positions(MSection):
         shape=[],
         unit='meter',
         description='''
-        -
         ''')
 
     x_orca_atom_positions_z = Quantity(
@@ -64,7 +61,6 @@ class x_orca_atom_positions(MSection):
         shape=[],
         unit='meter',
         description='''
-        -
         ''')
 
 
@@ -79,7 +75,6 @@ class x_orca_final_geometry(MSection):
         type=str,
         shape=[],
         description='''
-        -
         ''')
 
     x_orca_atom_positions_x_geo_opt = Quantity(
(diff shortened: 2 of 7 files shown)


ndaelman-hu:
Covers VASP, CASTEP, and Orca.

ndaelman-hu:
@JosePizarro3 Can I go ahead and merge?

JosePizarro3:
> @JosePizarro3 Can I go ahead and merge?

one second please, I was checking a couple of other things first, I will drop my comments

electronicparsers/castep/parser.py
```diff
@@ -626,6 +626,7 @@ def __init__(self):
             'Dispersion corrected final free energy* (Ecor-TS)': 'x_castep_total_dispersion_corrected_free_energy',
             'NB dispersion corrected est. 0K energy* (Ecor-0.5TS)': 'x_castep_disp_corrected_energy_total_T0',
             'basis set parameters': {
+                'basis set accuracy': 'native_tier',  # https://www.tcm.phy.cam.ac.uk/castep/documentation/WebHelp/content/modules/castep/keywords/k_basis_precision_castep.htm
```
JosePizarro3, line 629: maybe you can move all the online resources to the quantity def for `native_tier` (here and further down you have another link)

electronicparsers/orca/metainfo/orca.py
```diff
@@ -578,6 +571,17 @@ class Method(simulation.method.Method):
...
 
+class Scf(simulation.method.Scf):
+
+    m_def = Section(validate=False, extends_base_section=True)
+
+    x_orca_last_max_density_change = Quantity(
+        type=np.dtype(np.float64),
+        shape=[],
+        description='''
+        -
```
JosePizarro3: better to leave description empty than a hyphen

electronicparsers/orca/parser.py
```diff
@@ -599,6 +601,30 @@ def __init__(self):
...
+            'StrongSCF': 3.0e-07, 'TightSCF': 1.0e-08, 'VeryTightSCF': 1.0e-09,
+            'ExtremeSCF': 1.0e-14,
+        }  # https://sites.google.com/site/orcainputlibrary/numerical-precision
+        _native_tier_map = {k.lower(): v for k, v in _native_tier_map.items()}
+        sec_scf = Scf(
+            native_tier=self.out_parser.results['input_file']['tier'],
+        )
+        scf_convergence = self.out_parser.get('single_point', {}).\
+            get('self_consistent', {}).get('scf_convergence', {})
+        if threshold_energy_change := scf_convergence.get('last_energy_change',
```
JosePizarro3, line 616: Can this `:=` handle the potential error?

electronicparsers/orca/parser.py
```diff
@@ -599,6 +601,30 @@ def __init__(self):
...
+            trivial)[1]:
```
JosePizarro3: this will not pass pylint, I think.

```
if threshold_energy_change := scf_convergence.get(
    'last_energy_change', trivial)[1]
```

electronicparsers/orca/parser.py
```diff
@@ -599,6 +601,30 @@ def __init__(self):
...
+                sec_scf.threshold_energy_change = threshold_energy_change * ureg.hartree  # TODO: extract this from the setup, so it is present even whan a calcultion does not finish properly
+        sec_scf.threshold_density_change = scf_convergence.get('last_rms_density_change',
```
JosePizarro3, line 619: error handling here

electronicparsers/orca/parser.py
```diff
@@ -599,6 +601,30 @@ def __init__(self):
...
+            trivial)[1]
+        sec_scf.x_orca_last_max_density_change = scf_convergence.get(
```
JosePizarro3, line 621: error handling

electronicparsers/orca/parser.py
```diff
@@ -599,6 +601,30 @@ def __init__(self):
...
+            'last_max_density_change', trivial)[1]
+        if sec_scf.native_tier and not sec_scf.threshold_energy_change:
+            sec_scf.threshold_energy_change =\
```
JosePizarro3, line 624: is this `\` necessary? Cannot you leave it in the same line?

electronicparsers/orca/parser.py
```diff
@@ -599,6 +601,30 @@ def __init__(self):
...
+        _native_tier_map = {
+            'SloppySCF': 3.0e-05, 'LooseSCF': 1.0e-05, 'NormalSCF': 1.0e-06,
+            'StrongSCF': 3.0e-07, 'TightSCF': 1.0e-08, 'VeryTightSCF': 1.0e-09,
+            'ExtremeSCF': 1.0e-14,
+        }  # https://sites.google.com/site/orcainputlibrary/numerical-precision
+        _native_tier_map = {k.lower(): v for k, v in _native_tier_map.items()}
+        sec_scf = Scf(
+            native_tier=self.out_parser.results['input_file']['tier'],
+        )
+        scf_convergence = self.out_parser.get('single_point', {}).\
```
JosePizarro3, line 614: Can't this be left in the same line? Even if it is going a bit into the limit of tabs, I think for readability is better to keep it in the same line.

electronicparsers/orca/parser.py
```diff
@@ -599,6 +601,30 @@ def __init__(self):
...
 
+        trivial = [None] * 2
+        _native_tier_map = {
+            'SloppySCF': 3.0e-05, 'LooseSCF': 1.0e-05, 'NormalSCF': 1.0e-06,
+            'StrongSCF': 3.0e-07, 'TightSCF': 1.0e-08, 'VeryTightSCF': 1.0e-09,
+            'ExtremeSCF': 1.0e-14,
+        }  # https://sites.google.com/site/orcainputlibrary/numerical-precision
+        _native_tier_map = {k.lower(): v for k, v in _native_tier_map.items()}
+        sec_scf = Scf(
+            native_tier=self.out_parser.results['input_file']['tier'],
```
JosePizarro3: error handling here

electronicparsers/vasp/parser.py
```diff
@@ -619,6 +633,9 @@ def get_valence_basis_set(self) -> list[BasisSet]:
             sec_bases.append(sec_basis)
         return sec_bases
 
+    def get_tier(self, type='native') -> Union[str, None]:
+        return super().get_tier(self.parser.get('parameters').get('PREC'), type=type)
```
JosePizarro3: maybe this will produce some potential issue, but being in Outcar parsing I don't care, so you decide whether you want some error handling here

JosePizarro3 (commented):
Mostly error handlings, but not something that will take you too much.

Just one thing thought, as I saw the CASTEP and VASP testing, now `native_tier` contains similar things, e.g., same string but some code has it in capitals and the other in lower case. I'd say there should be some standarization there, and use `.lower()`. 

This links with my question about having a `MEnum` in the metainfo. I understand that you don't want to define it now, but at least we should be able to report the same strings for the same names, right?

electronicparsers/castep/parser.py
ndaelman-hu: These are **code-specific**, so I think they'll only be distracting there.
I agree that the current placing isn't optimal either, though. I briefly brought up links to code-specific documentation with Markus, but I see that I will have to follow up.

My proposal under the current setup would be to have a **section with links in the parser `yaml`**. What do you think?
I can run this past Markus or Alvin next week and then migrate all code lines (there are more) in the same issue where I rectify the linting. **What do you think of this proposal?**

electronicparsers/orca/metainfo/orca.py
ndaelman-hu: Okay, this was the style already present in the orca metainfo. I'll update all occurrences then.

electronicparsers/orca/parser.py
ndaelman-hu: I think that it is fine here (depends on which errors you mean), but let me explain a bit further, since this is a recent Python feature.

[The walrus operator](https://realpython.com/python-walrus-operator/) (`:=`) is just a trick to condense lines: it assigns `threshold_energy_change` and then immediately applies the `if`-condition.
The latter will catch empty statements. Since the `get` statements defaults to empty, this should cover any missing link in the chain (I know this from writing the tests).
I don't know which other runtime errors to expect here.

electronicparsers/orca/parser.py
ndaelman-hu: Pls see me comment on `:=`, above.

electronicparsers/orca/parser.py
ndaelman-hu: I'll see what pylint says.

electronicparsers/orca/parser.py
ndaelman-hu: I'll see what pylint says.

ndaelman-hu: It goes quite a bit over. In case of one-liners or more complex logic, I'd agree and use variables for better splitting.
Considering that the structure is quite repetitive here, I think it is clear enough though.

electronicparsers/orca/parser.py
JosePizarro3: Ah, great, I will then use it as well in my future coding.

Thanks for the info!

electronicparsers/orca/parser.py
JosePizarro3: but wait, this is not the same as in the comment of `:=` before. 

I'd say you could substitute all my "error handling" comments by `:=`?

electronicparsers/castep/parser.py
JosePizarro3: Being code-specific, I'll keep a few links on the description of the metainfo quantity itself (no need to actually cite ALL the codes, but just a few examples). I think you could see an example on how I envision it in `nomad/datamodel/metainfo/simulation/method.py:class BSE`, in `BSE.type`.

Your proposal might be a good idea, but maybe a bit overkill for us? At the end this is just a way of documenting knowledge for the future devs, so like you want; my opinion is that fewer links might be just enough.

electronicparsers/orca/parser.py
ndaelman-hu: But which error do you mean? This could only fail if `scf_convergence` is not set.

electronicparsers/castep/parser.py
ndaelman-hu: So you mean adding the links to the quantity, but not its description, right? This would be more to get a feeling of what is meant, correct?
In and of itself, the description should already be understandable. If we need links as devs, likelihood is that the same goes for the users.
My reasoning for the links here is that I want to justify a choice by referencing the manual. It's just to keep things fast and convenient, when the code might be obscuring the semantics.
I think that my suggestion is more appropriate for the latter goal. I'll bring this up with Markus, but for now keep it here. I'll migrate all of these refs, once we settle on a dev doc system (probably near future).

electronicparsers/orca/parser.py
JosePizarro3: No, this is a list you are then extracting the member [1]

ndaelman-hu: If you look at the definition `scf_convergence_quantities`, you'll see that there are always 2 floating point numbers extracted. Likewise, `trivial` is also a 2-member list. I can change the index to `-1`, that's somewhat more generic, but I don't think this warrant a catch for `IndexError`.

JosePizarro3: ah true, you are setting the default to `trivial`. Then indeed it should work.

I was scared of 'None' is not subscriptable (I think it is `TypeError` if I am not mistaken), which is very typical when you pass a `.get` and then try to resolve some member of a list.

PR #113, opened 2023-05-30
Native tier fhi-aims

ndaelman-hu:
Assign an overall `electrons_representation.native_tier` for FHI-aims. Note that FHI-aims has no catch-all keywords for regulating the calculation quality.
Rather, the code is shipped with some suggested settings (element-specific) that have to be copied over by the user.
Therefore, the `native_tier` is assigned based on other extracted (application) quantities.
Considering the range of customizability, the specific settings are read in from json files formerly in `nomad/normalizing`.


ndaelman-hu:
@ladinesa This is a first draft: I still have to extend the range of parameters and also add in tests.
Nonetheless, I'd like **your opinion on the overall design** I sketched out: both classes, their set of functionalities, as well as, the use of internal method calls.

ladinesa:
If I understand the code right, you would like to compare the parsed secition x_fhi_aims_section_controlIn_basis_set to the references in the json files and return the prefix of the filename matching the parsed values. In order to do this, you read the json file and store the data in a Storage object which you eventually use to do perform the comparison. For me this is a lot of code for a quantitiy. I think the Storage object is redundant as you can simply loop over all the sections and perform the comparison directly in the json file. If you are aiming for performance, I suggest we convert these files into one of the database formats already employed in nomad: archive or mongo. I suggest we use the nomad archive format as the data already follows the archive format. Please ask Theodore about how to create it and how to query.

ndaelman-hu:
> If I understand the code right

Yes, you've summarized it well! Let me just add that these JSON files were not set up by me, but are (unused) legacy.
They don't exactly follow the archive structure, but indeed mimic the `x_fhi_aims_section_controlIn_basis_set` section.
It also looks like some of the quantities changed mapping (previously having the quantity name end on and index and now a list).
I wrote `Storage` as a way to handle the reading and these minor conversions.

> If you are aiming for performance, I suggest we convert these files into one of the database formats already employed in nomad: archive or mongo.

You mean something like a set of fixed, FHI-aims database entries? I like this suggestion. It sounds like a platform that could be extended to other use cases. It would also allow us to work on tools for comparing entries. The only question would be _how to efficiently update then when the metainfo changes?_ I'll discuss with Theodore.

ladinesa:
We should certainly convert these files into msgpack files and we can use archive query functionality. We could go for a flat, schema-free database.

ndaelman-hu:
How do I convert the JSON to msgpack? I'll see to flatten the JSON a bit beforehand.

ladinesa:
> How do I convert the JSON to msgpack? I'll see to flatten the JSON a bit beforehand.

please have a look at nomad.archive.storage.write_archive then use nomad.archive.query

ndaelman-hu:
So, having discussed a couple of times with Theodore, given the small load on the RAM, he's suggesting different options:
1. Just load the JSON files into a single dictionary, which gets accessed. This would be faster at execution, but take 20 MB of disk space. Updating the JSON would also be easier.
2. Convert the JSON into MessagePack and keep the latter. It would give a larger reading overhead, but obviously decrease disk space.
The interface for writing the Archive, meanwhile, is too specialized for my needs. It is moreover set to be overhauled.

Theodore seemed most inclined towards option 1, given the scope atm. It is reasonable to expect a similar feature to be implemented for all parsers, eventually.

I guess most of the complexity in the code (me writing `Storage`, etc.) stems from the slightly awkward structure of these JSON files. If you want, I can write a small Jupyter Notebook, converting them to a single, more manageable structure. From there, the file would be loaded at _parser initialization_ and a single method would suffice for assigning the `native_tier`. If we want to preserve this conversion, I could also store it as a small git project.

@ladinesa What do you think of this approach?

ladinesa:
> So, having discussed a couple of times with Theodore, given the small load on the RAM, he's suggesting different options:
> 
> 1. Just load the JSON files into a single dictionary, which gets accessed. This would be faster at execution, but take 20 MB of disk space. Updating the JSON would also be easier.
> 2. Convert the JSON into MessagePack and keep the latter. It would give a larger reading overhead, but obviously decrease disk space.
>    The interface for writing the Archive, meanwhile, is too specialized for my needs. It is moreover set to be overhauled.
> 
> Theodore seemed most inclined towards option 1, given the scope atm. It is reasonable to expect a similar feature to be implemented for all parsers, eventually.
> 
> I guess most of the complexity in the code (me writing `Storage`, etc.) stems from the slightly awkward structure of these JSON files. If you want, I can write a small Jupyter Notebook, converting them to a single, more manageable structure. From there, the file would be loaded at _parser initialization_ and a single method would suffice for assigning the `native_tier`. If we want to preserve this conversion, I could also store it as a small git project.
> 
> @ladinesa What do you think of this approach?

i suggest we put all the files into a single json file following a flat structure so we do not have such a complicated query mechanism. We can maybe even create a hash for.each entry which we can compare if we have a uniform structure of the data.

ndaelman-hu:
@ladinesa I've compacted and flattened the tier references. The code for parsing is now vastly reduced.
Let me know:
1. if you agree with the approach
2. whether I need to convert it to MessagePack. Here, it seems that you and Theodore have opposite opinions.
3. what to do with the original files and the conversion script.
    Should they be added to their own Git project for perpetuity? I could use that project for composing other codes references too.

ladinesa:
> 2\. whether I need to convert it to MessagePack. Here, it seems that you and Theodore have opposite opinions.

no, i merely suggested using the archive format to leverage the query api in you are would like to consider optimizing performance. I am not initially aware of the size of the fhiaims data set so if plain json is better for this case then I have no problem with that.

> what to do with the original files and the conversion script. Should they be added to their own Git project for perpetuity? I could use that project for composing other codes references too.

For me, we do not need to keep the original files as we have all the data already in the modified file. The script, we also do not need  to keep it unless, it is a script which we can use to further extend the dataset then we can put it under utils but please make it a s python function. You can certainly put it in a separate github project and please put the link in utills.

electronicparsers/fhiaims/convert_tiers.ipynb
```diff
@@ -0,0 +1,246 @@
+{
```
ladinesa: please convert this into a python function, if we can use it to extend dataset. if this is only to convert the old dataset, I think we do not need it.

electronicparsers/fhiaims/parser.py
```diff
@@ -53,9 +54,20 @@
     x_fhi_aims_section_controlIn_basis_set, x_fhi_aims_section_controlIn_basis_func,\
     x_fhi_aims_section_controlInOut_atom_species, x_fhi_aims_section_controlInOut_basis_func,\
     x_fhi_aims_section_vdW_TS
+
 from ..utils import BeyondDFTWorkflowsParser
 
 
+json_file_name = 'native_tier_references.json'
```
ladinesa: put this under __init__. maybe register the name of the file in utils  then import it here. use NATIVE_TIER_REFERENCE_DATA_FILE. having filename also here is fine.

electronicparsers/fhiaims/parser.py
```diff
@@ -1575,6 +1587,30 @@ def parse_basis_set(species):
                 sec_basis_set.x_fhi_aims_controlIn_number_of_basis_func = len(division)
                 sec_basis_set.x_fhi_aims_controlIn_division = division
 
+        def _get_elemental_tier(basis_settings: x_fhi_aims_section_controlIn_basis_set,
```
ladinesa: i really like this much better, much simpler to maintain.

electronicparsers/fhiaims/parser.py
```diff
@@ -1575,6 +1587,30 @@ def parse_basis_set(species):
...
+                                reference: dict=native_tier_references) -> list[str]:
+            '''Compare the basis settings to the reference
+            and return the matching tier for each element.'''
+            tiers: list[str] = []
+            for setting in basis_settings:
```
ladinesa: one tiny thing, maybe put the loop for basis_settings outside in line 1687 to avoid these multiple indentations.

ndaelman-hu:
@JosePizarro3 I'm adding (close to finishing) a `native_tier` to FHI-Aims. Maybe you can take a quick look and tell me whether you foresee any conflicts with your workflows down the line. I honestly don't think so, but better safe than sorry.

electronicparsers/fhiaims/parser.py
ndaelman-hu: I originally tried this, but I get circular dependencies: `fhiaims/parser.py` now has to import from `fhiaims/__init__.py` and vice versa.
I could move the whole reading part to `utils`, yes, but that doesn't sound very appropriate: I thought that `uitls` is for general functionalities, across parsers.

electronicparsers/fhiaims/convert_tiers.ipynb
ndaelman-hu: This is only for converting the old dataset, yes.

electronicparsers/fhiaims/parser.py
ladinesa: I mean under FHIAimsParser.__init__

JosePizarro3:
> @JosePizarro3 I'm adding (close to finishing) a `native_tier` to FHI-Aims. Maybe you can take a quick look and tell me whether you foresee any conflicts with your workflows down the line. I honestly don't think so, but better safe than sorry.

It does not look like to me. You can of course try the test files to make sure. Thanks!

electronicparsers/fhiaims/parser.py
ndaelman-hu: Like this? I'm still not sure what you meant by importing from `utils`, but I guess this is also fine.

electronicparsers/fhiaims/parser.py
```diff
@@ -886,6 +884,24 @@ def __init__(self):
...
+        _package_folder = os.path.dirname(__file__)
+        _native_tier_reference_data_filepath = os.path.join(_package_folder,
+            _native_tier_reference_data_filename)
+        try:
+            with open(_native_tier_reference_data_filepath) as f:
+                self._native_tier_references = json.load(f)
+        except (FileNotFoundError, json.JSONDecodeError):
+            raise Exception(f'''
+                FHI-aims parser could not load {_native_tier_reference_data_filename}
+            ''')  # TODO: add to error logger
```
ndaelman-hu: @ladinesa should I apply the Error Logger here?

electronicparsers/fhiaims/parser.py
ladinesa: My idea was to put all global variables in some config file. but having it here for the moment is also fine.

electronicparsers/fhiaims/parser.py
ladinesa: please do not put a raise. put a warning instead and return None.

electronicparsers/fhiaims/parser.py
```diff
@@ -880,6 +882,20 @@ def __init__(self):
         # max cumulative number of atoms for all parsed trajectories to calculate sampling rate
         self._cum_max_atoms = 100000
 
+        # set up the native tier references
+        self._tier_map = {None: -1, 'light': 1, 'tight': 2, 'really_tight': 3}
```
ladinesa: sorry just noticed it now, do we need this map or we can put it also in the dictionary?

electronicparsers/fhiaims/parser.py
```diff
@@ -880,6 +882,20 @@ def __init__(self):
...
+        _native_tier_reference_data_filename = 'native_tier_references.json'
+        _package_folder = os.path.dirname(__file__)
```
ladinesa: not necessary to define _package_folder

electronicparsers/fhiaims/parser.py
```diff
@@ -1552,6 +1568,26 @@ def parse_basis_set(species):
                 sec_basis_set.x_fhi_aims_controlIn_number_of_basis_func = len(division)
                 sec_basis_set.x_fhi_aims_controlIn_division = division
 
+        def _get_elemental_tier(basis_settings: x_fhi_aims_section_controlIn_basis_set,
+                reference: dict=self._native_tier_references) -> list[str]:
+            '''Compare the basis settings to the reference
+            and return the matching tier for each element.'''
+            tier_refs = reference['data'][basis_settings['x_fhi_aims_controlIn_species_name']]
+            for tier_name, tier_ref in tier_refs.items():
```
ladinesa: do you think we can go for a sort of hash instead of comparing the items?

ndaelman-hu: Potentially, but we'd need to filter out a couple of quantities from `x_fhi_aims_section_controlIn_basis_set` before hashing the NOMAD version.
Atm, Luca requested me to make a simple, direct comparison (`==`). If we stick to that, a hash is probably the best choice, yes.
Imo, an inequality (`<`) makes more sense though, since we want to build out a scale. I'd work towards catching something like `light_2020` but with some more fine-tuning as `light_2020+`.

electronicparsers/fhiaims/parser.py
ndaelman-hu: Short version: I don't think so

Long version:
`native_tier` is supposed to reflect the whole calculation. In FHI-aims, one can set the precision by element, though.
I don't think that users want to bother further refining their search, so I'd only give them the roughest result.
This map is for assigning exactly that. The dictionary, meanwhile, specifies the settings of each element.

electronicparsers/fhiaims/parser.py
ndaelman-hu: Okay, this would be something like `fhiaims/utils.py`? Could we leave this as a refactoring issue?

ladinesa: electronicparsers/config.py, but yes we can do that in a separate issue together with the others.

electronicparsers/fhiaims/parser.py
ndaelman-hu: If (for the time being) I were to go with a hash, is there a standard that you use?
What are some aspects to watch out for when picking a hash function? The amount of independent variables?

ladinesa: you can look at hashlib. we can also use built in hash(sorted(frozenset(data)))

electronicparsers/fhiaims/parser.py
```diff
@@ -1552,6 +1573,45 @@ def parse_basis_set(species):
...
+            to_be_filtered = [prefix + k for k in to_be_filtered]
+            basis_new = {}
+            if not isinstance(basis_settings, dict):
+                basis_settings = basis_settings.m_to_dict()
+            for k, v in basis_settings.items():
+                if (k.startswith(prefix) or k.startswith(prefix_repeating)) and k not in to_be_filtered:
+                    basis_new[k] = v
+            return {k: basis_new[k] for k in sorted(basis_new.keys())}
+
+        def _get_elemental_tier(basis_settings: x_fhi_aims_section_controlIn_basis_set,
```
ladinesa: it seems that the reference argument is not necessary.

ndaelman-hu: I set it to a default, just in case it ever has to be changed

electronicparsers/fhiaims/parser.py
ndaelman-hu: I applied a hash with sorted over a `dict`

electronicparsers/fhiaims/parser.py
ndaelman-hu: Has been put into the JSON reference file

ndaelman-hu:
While it technically works now (at least for the examples tested, but this should extrapolate quite well), it's running quite slow.
I'm not too happy with the performance there. I'd have to dig deeper to find the root cause. Thusfar, I can only tell that it doesn't affect the timing of the other tests. 

@ladinesa Could I leave this for a subsequent PR?

electronicparsers/fhiaims/parser.py
```diff
@@ -1552,6 +1573,45 @@ def parse_basis_set(species):
                 sec_basis_set.x_fhi_aims_controlIn_number_of_basis_func = len(division)
                 sec_basis_set.x_fhi_aims_controlIn_division = division
 
+        def _prep_elemental_tier(basis_settings: MSection) -> dict[str, Any]:
```
ladinesa: I would put this under _get_elemental_tier, if not used anywhere.

ndaelman-hu:
@ladinesa Could you please exclude test data from the linting tests?

ladinesa:
> @ladinesa Could you please exclude test data from the linting tests?

please update .pylintrc

ladinesa:
> While it technically works now (at least for the examples tested, but this should extrapolate quite well), it's running quite slow. I'm not too happy with the performance there. I'd have to dig deeper to find the root cause. Thusfar, I can only tell that it doesn't affect the timing of the other tests.
> 
> @ladinesa Could I leave this for a subsequent PR?

I would deal with it here before we merge. I can help you profile if you like.

ndaelman-hu:
> > While it technically works now (at least for the examples tested, but this should extrapolate quite well), it's running quite slow. I'm not too happy with the performance there. I'd have to dig deeper to find the root cause. Thusfar, I can only tell that it doesn't affect the timing of the other tests.
> > @ladinesa Could I leave this for a subsequent PR?
> 
> I would deal with it here before we merge. I can help you profile if you like.

Thx for the offer. Luca asked me to have this for the tutorial (it's the last of these additional features), so I still want to give Markus time for process these.

electronicparsers/fhiaims/parser.py
```diff
@@ -1552,6 +1573,45 @@ def parse_basis_set(species):
...
+            bs_filtered[orbital_name] = [_prep_elemental_tier(orb) for orb in bs_filtered[orbital_name]]
+            for quantity in ['radius', 'type', 'l', 'n']:
+                try:
+                    bs_filtered[orbital_name] = sorted(bs_filtered[orbital_name], key=lambda x: x[f'x_fhi_aims_controlIn_basis_func_{quantity}'])
+                except KeyError:
+                    pass
+            bs_hash = hashlib.sha1()
+            bs_hash.update(json.dumps(bs_filtered, sort_keys=True).encode('utf-8'))
+            for tier in reference['hash'][identifier]:
+                if bs_hash.hexdigest() == reference['hash'][identifier][tier]:
```
ladinesa: i think hexdigest should be outside the loop

ladinesa: Also, the hash should be the key for the so we do not do loop but we simply access it by reference[bs_hash.hexdigets()]

ndaelman-hu: Agreed! I'll see to apply this

electronicparsers/fhiaims/parser.py
ladinesa: I suggest we make it as a flattened dict, so we do not have to filter the elements. Also do not separate data and hierarchy from hash so we do not perform unnecessary mapping.

electronicparsers/fhiaims/parser.py
```diff
@@ -1552,6 +1573,47 @@ def parse_basis_set(species):
...
+            # filter out element identifiers and other non-relevant quantities
+            id_name = 'x_fhi_aims_controlIn_species_name'
+            orbital_name = 'x_fhi_aims_section_controlIn_basis_func'
+            if (identifier := basis_settings[id_name]) is None:
+                return
+            bs_filtered = _prep_elemental_tier(basis_settings)
+            bs_filtered[orbital_name] = [_prep_elemental_tier(orb) for orb in bs_filtered[orbital_name]]
+            for quantity in ['radius', 'type', 'l', 'n']:
+                try:
+                    bs_filtered[orbital_name] = sorted(bs_filtered[orbital_name], key=lambda x: x[f'x_fhi_aims_controlIn_basis_func_{quantity}'])
```
ladinesa: line break

electronicparsers/fhiaims/parser.py
ndaelman-hu: There are some interesting ideas here, but I get the feeling that others are getting conflated.
- The filtering and flattening don't have anything to do with each other. The filtering was just to cut down to the essentials quantities to compare. I can also see to add them to the JSON. I wouldn't do so, though, since I'm unsure that the hashes will stay.
- The only section to flatten would be `x_fhi_aims_section_controlIn_basis_func`, but it is repeated... Flattening it, would entail changing the keyword structure, which will make the object harder to reason about. For the same reason, the code in the parser would also become bigger. Do you see another option?

> Also do not separate data and hierarchy from hash so we do not perform unnecessary mapping.

I'll incorporate `hierarchy` into `hash`, thank you for the suggestion. `data` is not really used, it's just there to explain what the hash encrypts. This is irrelevant for the parsing itself, but hugely important for development and comparison purposes. Maybe it's better to separate both into separate files / projects (e.g. https://github.com/FAIRmat-NFDI/AreaC-Data_Quality_AbInitio), with a small `info` section explaining the link? **I do insist that `data` in one way or shape should remain** (sha1 cannot be unhashed, and even then we need something more readable).

ladinesa: Ah thanks for clarifying. I thought you still require some more info from data. So indeed we need to store it separately as reference. I hope this will address the performance issue. We can then have any format for data.

ndaelman-hu: Sure thing! I implemented the suggestion above. There's still one test that runs slow, not sure what's causing this.
Considering the timeframe of the tutorial, are you fine with me merging and optimizing later?

PR #116, opened 2023-06-08
Fix bad unrolling on the exciting APW basis set

ndaelman-hu:
Sebastian Tillack from the SOL group pointed out to me that assignments of the APW wavefunctions was flawed.
I traced the problem back to bugs in the extraction of the `<custom>` and `<lo>` tags.
I also used this opportunity to clean up the code.

### DIFF
--- a/electronicparsers/exciting/parser.py
+++ b/electronicparsers/exciting/parser.py
@@ -1622,7 +1622,7 @@ class ExcitingParser(BeyondDFTWorkflowsParser):
                         ]
                     )
                 ]
-            sec_method.electrons_representation[0].basis_set.append(bs_val)
+        sec_method.electrons_representation[0].basis_set.append(bs_val)
 
     def parse_file(self, name, section, filepath=None):
         # TODO add support for info.xml, wannier.out
--- a/electronicparsers/exciting/parser.py
+++ b/electronicparsers/exciting/parser.py
@@ -109,7 +109,7 @@ class SpeciesParser(TextParser):
                 'atomicState', rf'<atomicState{re_tags}/?>',
                 sub_parser=TextParser(quantities=[q_key_val]), repeats=True),
             Quantity(
-                'default', r'(<default[\s\S]+(</default>|/>))',
+                'default', r'(<default[\s\S]+?(</default>|/>))',
                 sub_parser=TextParser(quantities=[q_wf, q_key_val])),
             Quantity(
                 'custom', r'(<custom[\s\S]+?(</custom>|/>))',
@@ -1572,12 +1572,15 @@ class ExcitingParser(BeyondDFTWorkflowsParser):
             radius_lin_spacing=radial_spacing * ureg.bohr,
         )
         lo_samplings = {lo['l']: lo['wf'] for lo in species_data.get('lo', [])}
+        orbital_type_default = species_data['default']['type']
+        if orbital_type_default is None:
+            orbital_type_default = 'lapw'
+        energy_parameter_default = species_data['default']['trialEnergy']
         lmax = self.input_xml_parser.get('xs/lmaxapw', 10)
+
         for l_n in range(lmax + 1):
-            orbital_type = species_data['default']['type']
-            energy_parameter = species_data['default']['trialEnergy']
-            if orbital_type is None:
-                orbital_type = 'lapw'
+            orbital_type = orbital_type_default
+            energy_parameter = energy_parameter_default
             for custom_settings in species_data.get('custom', []):
                 if custom_settings['l'] == l_n:
                     orbital_type = custom_settings['type']
--- a/electronicparsers/exciting/parser.py
+++ b/electronicparsers/exciting/parser.py
@@ -126,15 +126,15 @@ class SpeciesParser(TextParser):
             except IndexError:
                 return key_vals
 
-        def _supplant(val_map: dict[Any, Any], flag: str = 'key_val') -> dict:
-            supplanted = {}
+        def _supplant(val_map: dict[str, Any], flag: str = 'key_val') -> dict[str, Any]:
+            supplanted: dict[str, Any] = {}
             for k, v in val_map.items():
                 if k == flag:
                     return _convert_key_val(v)
                 elif isinstance(v, dict):
                     supplanted[k] = _supplant(v)
                 elif isinstance(v, Iterable):
-                    supplatend_v: list[dict[Any, Any]] = []
+                    supplatend_v: list[dict[str, Any]] = []
                     for vv in v:
                         if isinstance(vv, dict):
                             supplatend_v.append(_supplant(vv))
--- a/electronicparsers/exciting/parser.py
+++ b/electronicparsers/exciting/parser.py
@@ -1598,19 +1598,17 @@ class ExcitingParser(BeyondDFTWorkflowsParser):
 
             # Add lo's
             max_order = 2
-            if orbital_type[-2] == 'LO' and l_n in lo_samplings.keys():
-                energy_parameters = [energy_parameter for _ in range(max_order)]
+            if orbital_type[-2:] == 'LO' and l_n in lo_samplings:
                 for wf in lo_samplings[l_n]:
-                    energy_parameters = [wf['trialEnergy'] for _ in range(wf['matchingOrder'], max_order)]
-                    for order, energy_parameter in enumerate(energy_parameters):
+                    for order in range(wf.get('matchingOrder', 0), max_order):
                         orbital = OrbitalAPW(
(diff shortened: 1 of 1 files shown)


ndaelman-hu:
@ladinesa If Sebastian replies in time, I could also add a test here.

electronicparsers/exciting/parser.py
```diff
@@ -109,49 +109,44 @@ def init_quantities(self):
...
-                    for v in val:
-                        if isinstance(v, TextParser):
-                            self._run_tree(v)
-        return parser
-
-    def parse(self, key=None):
-        self = super().parse(key=key)
-        self = self._run_tree(self)
-        return self
+    def to_dict(self) -> dict:
```
ladinesa, line 122: FileParser has already a method to_dict. This recursively goes through all the sub parsers and convert the contents to dict. I am not sure why you need to convert it to dict. I was looking at the species_parser and it seems that it is not necessary. Most probably I am not understanding what is going on here. I think we can get rid of these unnecessary complications if we use the xml parser.

ndaelman-hu: The line `q_key_val = Quantity(self.flag, r'([a-zA-Z]+)\="([\w\.\-\+]+)"', repeats=True)` generates (after `FileParser.to_dict()`) a structure like `{key_val: [['l', 1], ['type', 'apw+lo'], ['trialEnergy', 0.15], ['searchE', 'false']]}`.
I would like to have it as `{'l': 1, 'type': 'apw+lo', 'trialEnergy': 0.15, 'searchE': 'false'}`

electronicparsers/exciting/parser.py
```diff
@@ -1557,10 +1552,21 @@ def _parse_band_out(self, sec_scc):
             sec_k_band_segment.value = band_energies[nb] + energy_fermi
 
     def _parse_species(self, sec_method):
-        type_order_mapping = {'APW': 1, 'LAP': 2}
+        def _set_orbital(source: TextParser, l_quantum_number: int,
```
ladinesa, line 1553: not necessary to define this. it is used twice, but the function is simple and you would supply most parameters.

ndaelman-hu: `type_order_mapping` you mean?

electronicparsers/exciting/parser.py
```diff
@@ -92,7 +92,7 @@ def __init__(self):
 
     def init_quantities(self):
```
ladinesa, line 93: I suggest you use the xml parser or make a custom xml parser. Although TextParser can handle all types of files, I think performance-wise python xml module is better.

ndaelman-hu: I see... It definitely crossed my mind, but I went with the `TextParser` due to the amount of tag attributes and the file's small size.
At this point, changing it would warrant quite a rewrite for something that's not a bottleneck. Are you fine if I keep the `TextParser` here?

ladinesa: Sure, I will create an issue so we can continue working on it.

electronicparsers/exciting/parser.py
ladinesa: No, _set_orbital, though it is a matter of taste really.

electronicparsers/exciting/parser.py
ladinesa: Ah, but is a one-liner function not enough for this conversion? We can add a list_to_dict argument to Quantity but this would somehow break the repeating nature  and type of the Quantity

electronicparsers/exciting/parser.py
ndaelman-hu: Feel free to assign it to me.

electronicparsers/exciting/parser.py
ndaelman-hu: There are 2 places where the orbital is constructed, mostly using the same structure.
Just wanted to leverage that and cut down on code. I'll let it be then.

electronicparsers/exciting/parser.py
ndaelman-hu: Essentially, yes, that's what `_convert_key_val` is for. As you mentioned, it depends on whether I can pass this to the `Quantity`. Such a feature goes quite deep in the feature set, so I prefer consulting you before adding it. It would indeed change the repeating nature, but one would know due to the attribute.
 
The more verbose part of `SpeciesParser.to_dict` just manages the unpacking of the `dict`, since this is now done after `to_data` and `to_dict`.

Should we try to get this feature in now, or maybe better that I open a new issue and we also involve Markus in the discussion there.

ndaelman-hu:
@ladinesa Got a question: the updates to the `exciting` parser show up in the JSON output of the archive. However, when I parse with appworker, it doesn't show up in the GUI, neither locally, nor when deployed. No modifications were made to the metainfo.

Do you have an idea of where the problem could be?

electronicparsers/exciting/parser.py
ladinesa: I personally would not include this as a quantity method and would put it in the parser. I still do not understand why the to_dict is overly complicated. Can we not convert when we need to? I am just worried about the performance impact of the for loops.

ladinesa:
> Do you have an idea of where the problem could be?

Maybe it is not using the correct parser commit or that you did not install using -e

ndaelman-hu:
> Maybe it is not using the correct parser commit or that you did not install using -e

 Triple-checked the commit. I always install with `-e` and that wouldn't explain why the deployment fails too:
https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1342?commit_id=8b2b4e28c2d0e7bd9dc60bad27241b267672066e

electronicparsers/exciting/parser.py
ndaelman-hu: > I personally would not include this as a quantity method

I could see it being useful in other occasions too: there will be situations where the keyword and value can be matched together.
The more natural way to interact with them would then be this kind of structure.

ladinesa:
> > Maybe it is not using the correct parser commit or that you did not install using -e
> 
> Triple-checked the commit. I always install with `-e` and that wouldn't explain why the deployment fails too: https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1342?commit_id=8b2b4e28c2d0e7bd9dc60bad27241b267672066e

maybe put a raise in the parser to see if the worker picks it up.

electronicparsers/exciting/parser.py
ladinesa: It is way too specific as it only works for key-val pairs and as i have said, you can do it in one line.

ndaelman-hu: How would you do the oneliner after `to_dict`?

ladinesa: If you encounter your tag then convert. Anyway, we can let this be, I think we can simplify this when we use the xml parser. Sorry for being too picky but I am just worried about performance and these small things add up.

ndaelman-hu:
> maybe put a raise in the parser to see if the worker picks it up.

Okay, traced it back to `ExcitingParser.file_exists()`: it doesn't find any species files when run from appworker, but it does as a local parser

ladinesa:
> > maybe put a raise in the parser to see if the worker picks it up.
> 
> Okay, traced it back to `ExcitingParser.file_exists()`: it doesn't find any species files when run from appworker, but it does as a local parser

I was just looking at this earlier.There is a utils function get_files. Maybe we can repurpose this.

ndaelman-hu:
> I was just looking at this earlier.There is a utils function get_files. Maybe we can repurpose this.

I basically just need it to accept regexes too. I guess that I'm misunderstanding something about the message pack file system and this is why`file_exists` (with `fuzzy=True`) fails

ndaelman-hu:
@ladinesa Solved the problem locally, but not on the deployment... Could you please take a look?

ndaelman-hu:
> @ladinesa Solved the problem locally, but not on the deployment... Could you please take a look?

Just needed time, all good now

electronicparsers/exciting/parser.py
ndaelman-hu: No prob, I can open an issue for optimization.
Can this one be closed then? No feedback yet, so I'm holding off on implementing a test.

ladinesa: > No prob, I can open an issue for optimization. Can this one be closed then? No feedback yet, so I'm holding off on implementing a test.

If all tests pass you can merge. Also, if you have time please do some benchmarking. thanks

PR #117, opened 2023-06-08
Improve get_mainfile_keys peformance

ladinesa:


JosePizarro3:
Ok great, I was not aware of `findall` to be set to False. I was checking a bit `TextParser` and I think I understand the difference between `re.findall` and `re.finditer`. 

Feel free to merge at will, and thanks for the fix.

JosePizarro3 (approved):

Issue #122, opened 2023-06-09
POTCAR.stripped not parsed


ndaelman-hu:
The `PotParser` picks up output from the `OUTCAR`, but not `POTCAR.stripped`

PR #123, opened 2023-06-09
122 Parse alternate version of POTCAR.stripped

ndaelman-hu:
Encountered an example of alternate version of POTCAR.stripped (now added to test data), where the pseudopotential sections end differently.


ndaelman-hu:
closes #122

PR #129, opened 2023-06-26
128 vasp hubbard parameters wrongly cast

### LINKED ISSUE #128: VASP: Hubbard parameters wrongly cast
Hubbard parameters LDAU, LDAUL, LDAUU, LDAUJ extracted from vasprun.xml of INCAR are cast to integers or floats, not lists, when only 1 entry is present. The later mapping to the NOMAD schema assumes a list, though. Both should be better attenuated to each other.

TO DO: double-check OUTCAR extraction of the same values

ndaelman-hu:
closes #128


markus1978:
Looks good to me.

PR #135, opened 2023-07-20
Fix reading booleans in VASP incar

lauri-codes:


lauri-codes:
@ladinesa: There was a problem with parsing boolean values such as `.FALSE.` and `.TRUE.` Does this fix look reasonable to you or did I miss something?

ladinesa:
looks right for me. thanks for the fix.

PR #139, opened 2023-07-21
126 apw missing energy parameters

### LINKED ISSUE #126: APW missing energy parameters
The `energy_parameter` in the APW codes is reported as an `undefined vector`. This has been verified for exciting and Wien2k, whereas fleur does not apply.

ndaelman-hu:
This is an intermediate solution. It adds the missing energy parameters, corrects the FP-APW parsing in exciting and already prepares for the next release.
It also provides some general functionalities for _unrolling_ all APW orbitals.

TODO: 
1. Extend to Wien2k and Fleur (maybe Elk?)
2. Add testing for the general functionalities and the new exciting version (waiting for examples).

Closes #126

### DIFF
--- a/electronicparsers/exciting/parser.py
+++ b/electronicparsers/exciting/parser.py
@@ -27,7 +27,7 @@ from nomad.datamodel.metainfo.simulation.run import Run, Program
 from nomad.datamodel.metainfo.simulation.method import (
     Method, DFT, Electronic, Smearing, XCFunctional, Functional, Scf, BasisSet, KMesh,
     FrequencyMesh, Screening, GW, Photon, BSE, CoreHole, BasisSetContainer,
-    OrbitalAPW, AtomParameters,
+    AtomParameters,
 )
 from nomad.datamodel.metainfo.simulation.system import (
     System, Atoms
@@ -51,7 +51,7 @@ from .metainfo.exciting import (
     x_exciting_scrcoul_parameters
 )
 from ..utils import (
-    get_files, BeyondDFTWorkflowsParser
+    get_files, BeyondDFTWorkflowsParser, OrbitalAPWConstructor,
 )
 from typing import Any, Iterable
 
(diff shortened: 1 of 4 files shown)


electronicparsers/exciting/parser.py
```diff
@@ -1550,77 +1550,130 @@ def _parse_band_out(self, sec_scc):
...
+            return
+
+        # helper functions
+        def _convert_keyval(key_vals: list[list]) -> dict[str, Any]:
+            '''Helper function to convert key-value pairs to a dictionary'''
+            bool_map = {'false': False, 'true': True}
+            key_vals_converted: dict[str, Any] = {}
+            for key_val in key_vals:
+                if len(key_val) != 2:
+                    raise ValueError(f'Invalid key-value pair: {key_val}')
```
ladinesa, line 1580: replace raise with a warning.

electronicparsers/utils/utils.py
```diff
@@ -283,3 +286,136 @@ def extract_polarization_outputs():
             workflow.m_add_sub_section(ParticleHoleExcitations.tasks, task)
 
         xs_workflow_archive.workflow2 = workflow
+
+
+class OrbitalAPWConstructor:
```
ladinesa, line 291: please add test

electronicparsers/utils/utils.py
```diff
@@ -283,3 +286,136 @@ def extract_polarization_outputs():
...
+        if self.wavefunctions:
+            for key, val in self.wavefunctions.items():
+                converted = self._convert(settings, **kwargs)
+                if getattr(OrbitalAPW, key, {}).get('shape'):
+                    try:
+                        self.wavefunctions[key].append(converted[key])
+                    except AttributeError:
+                        self.wavefunctions[key] = np.append(self.wavefunctions[key], converted[key])
+                elif val != converted[key]:
+                    raise ValueError(f'Wavefunction {key} does not match previous value')
```
ladinesa, line 351: avoid raising exceptions

electronicparsers/utils/utils.py
```diff
@@ -283,3 +286,136 @@ def extract_polarization_outputs():
...
+
+    def _convert(self, settings: dict[str, Any], **kwargs) -> dict[str, Any]:
+        '''
+        Convert the input arguments (`args`) to a `dict` as defined by `input_format`.
+        Keys outside of `input_format` can be added via kwargs.
+        '''
+        return dict(sorted({**settings, **kwargs}.items()))
+
+    def append_wavefunction(self, settings, **kwargs):
+        '''
```
JosePizarro3, line 332: Can you please add descriptions here and below in other functions?

electronicparsers/utils/utils.py
```diff
@@ -283,3 +286,136 @@ def extract_polarization_outputs():
...
+        '''
+        if self.wavefunctions:
+            for key, val in self.wavefunctions.items():
+                converted = self._convert(settings, **kwargs)
+                if getattr(OrbitalAPW, key, {}).get('shape'):
```
JosePizarro3: @ndaelman-hu @ladinesa is this correct? I remember Theodore saying that one can safely use getattr for MSections, but just wanted to make sure 🙂

electronicparsers/utils/utils.py
```diff
@@ -283,3 +286,136 @@ def extract_polarization_outputs():
...
+
+        # sort out the orbitals by id_keys
+        for term in self.term_order:
+            orbitals = sorted(self.orbitals, key=lambda orbital: _sort_func(orbital, term))
+        formatted_orbitals: list[OrbitalAPW] = []
+        for orbital in orbitals:
+            # convert to NOMAD section
+            formatted_orbital = OrbitalAPW()
+            for term, val in orbital.items():
+                setattr(formatted_orbital, term, val)
```
JosePizarro3, line 444: Coming back to my comment above on `getattr` (which again, it is just to confirm, so most probably you can use it) `setattr` should not be used and instead you should `m_add_sub_section`. 

You have an example in `nomad/normalizing/method.py:ExcitedStateMethod`, at the end I was adding to simulation depending on the dict `self._method_def` (which is defined at the beginning of the `MethodNormalizer()`.

electronicparsers/utils/utils.py
```diff
@@ -283,3 +286,136 @@ def extract_polarization_outputs():
...
+                'energy_parameter_n', 'energy_parameter', 'order', 'updated',
+            ]
+        )
+        self.term_order.reverse()
+        self.comparison_keys = kwargs.get(
+            'comparison',
+            ['l_quantum_number', 'j_quantum_number', 'k_quantum_number']
+        )
+
+    def _convert(self, settings: dict[str, Any], **kwargs) -> dict[str, Any]:
```
JosePizarro3, line 324: I think this could actually live out of `OrbitalAPWConstructor`, but you decide. Furthermore, the naming of the function could be more verbose: `_convert_to_dict`.

electronicparsers/utils/utils.py
```diff
@@ -283,3 +286,136 @@ def extract_polarization_outputs():
...
+            if orbital_ids != converted_ids:
+                new_orbitals.append(orbital)
+        self.orbitals = new_orbitals
+        self.append_orbital(*settings, **kwargs)
+
+    def get_orbitals(self) -> list[OrbitalAPW]:
+        '''
+        Return the stored orbitals as a sorted list of `OrbitalAPW` sections.
+        '''
+        def _sort_func(orbital: dict[str, Any], term: str):
```
JosePizarro3, line 423: I am not sure what this function does.

JosePizarro3 (changes requested):
Ok, finished with my review.

I can only comment on the fact that `OrbitalAPWConstructor` is not very understandable right now. Do you have any examples at hand that you can share? (I think that's why Alvin asked you for testing on this).

Furthermore, I would be more verbose. Especially with names of functions, descriptions, names of variables, etc. You can also add comments to make it even more clear.

Besides these comments, do you mind checking the `getattr` / `setattr` thingy? Also, pycodestyle is failing.

ndaelman-hu:
> I can only comment on the fact that OrbitalAPWConstructor is not very understandable right now.

Indeed, documentation is something that I still wanted to add. `OrbitalAPWConstructor` is not necessarily finished either. It was only applied to _exciting_ and might need some additional changes. Basically, its purpose is as a toolkit for handling different schemas on how the APW basis is written by code. Orbitals are defined for each l-channel (quantum angular momentum), but often times it is written in a more compressed format by the code. Typically, these will have some kind of default to which they add or amend orbitals. Instead, in NOMAD, I am showing them in a more verbose or "unrolled" way (each compression is different).

I am generating some figures to aid in the explanation:
apw_unrolling

electronicparsers/exciting/parser.py
ndaelman-hu: This means that the basis_set output will be complete bull. When should an error be fatal?

ladinesa: but we do not want to terminate parsing due to this. simply do not write to archive if deemed incorrect.

electronicparsers/utils/utils.py
ladinesa: getattr is fine, just avoid setattr.

electronicparsers/utils/utils.py
ladinesa: you can do formatted_orbital.m_set(formatted_orbital.m_get_quantity_defintion(tem), val)

PR #141, opened 2023-07-31
Fix CP2K SinglePoint

JosePizarro3:
Hi @ladinesa ,

I wanted to fix the SinglePoint as well as adding the DOS parsing to CP2K, but the later is going to be much more involved. Thus I simply push now the fix, and then work in a separate branch on parsing the DOS.

Can you please review it?

### DIFF
--- a/electronicparsers/cp2k/parser.py
+++ b/electronicparsers/cp2k/parser.py
@@ -1293,6 +1293,8 @@ class CP2KParser:
                     self.parse_system(atomic_coord)
                 else:
                     self.parse_system(n)
+                if sec_run.system[n] is not None:
+                    sec_scc.system_ref = sec_run.system[n]
 
         single_point = quickstep.get('single_point')
         if single_point is not None:
@@ -1555,7 +1557,7 @@ class CP2KParser:
                 self._calculation_type = calculation_type
                 break
 
-        sec_run = self.archive.m_create(Run)
+        self.archive.m_create(Run)
         self.parse_settings()
         self.parse_input()
 
--- a/electronicparsers/cp2k/parser.py
+++ b/electronicparsers/cp2k/parser.py
@@ -1071,13 +1071,13 @@ class CP2KParser:
                         val = int(val.strip('svn:'))
                     except Exception:
                         continue
-                setattr(sec_program_information, 'x_cp2k_%s' % key, val)
+                sec_program_information.m_set(sec_program_information.m_get_quantity_definition(f'x_cp2k_{key}'), val)
 
         dbcsr_settings = self.settings.get('dbcsr', {})
         if dbcsr_settings:
             sec_dbcsr = sec_run.m_create(x_cp2k_section_dbcsr)
             for key, val in dbcsr_settings.items():
-                setattr(sec_dbcsr, 'x_cp2k_%s' % key, val)
+                sec_dbcsr.m_set(sec_dbcsr.m_get_quantity_definition(f'x_cp2k_{key}'), val)
 
         program_settings = self.settings.get('program', {})
         if program_settings:
@@ -1090,13 +1090,13 @@ class CP2KParser:
                     key, val = 'start_id', val[0]
                 section = sec_endinformation if key.startswith('end') else sec_startinformation
                 val = val[0] if isinstance(val, list) else val
-                setattr(section, 'x_cp2k_%s' % key, val)
+                section.m_set(section.m_get_quantity_definition(f'x_cp2k_{key}'), val)
 
         global_settings = self.settings.get('global', {})
         if global_settings:
             sec_global_settings = sec_run.m_create(x_cp2k_section_global_settings)
             for key, val in global_settings.items():
-                setattr(sec_global_settings, 'x_cp2k_%s' % key, val)
+                sec_global_settings.m_set(sec_global_settings.m_get_quantity_definition(f'x_cp2k_{key}'), val)
 
         restart = self.out_parser.get('restart')
         if restart is not None:
@@ -1132,8 +1132,7 @@ class CP2KParser:
                 if sec_def is not None:
                     sub_section = section.m_create(sec_def.section_cls)
                     for key, val in data.items():
-                        sec_name = '%s_%s' % (name, key)
-                        parse(sec_name, val, sub_section)
+                        parse(f'{name}_{key}', val, sub_section)
 
             elif isinstance(data, list) and data:
                 for val in data:
@@ -1144,7 +1143,7 @@ class CP2KParser:
                 name = override_keyword(name)
                 quantity_def = resolve_definition(name)
                 if quantity_def is not None:
-                    setattr(section, name, quantity_def.type(data))
+                    section.m_set(section.m_get_quantity_definition(name), quantity_def.type(data))
 
         input_files = get_files(input_filename, self.filepath, self.mainfile, deep=False)
         if not input_files:
@@ -1189,7 +1188,7 @@ class CP2KParser:
                         sec_scf_energy.m_add_sub_section(getattr(
                             Energy, key.replace('energy_', '')), EnergyEntry(value=val))
                     else:
-                        setattr(sec_scf, key, val)
+                        sec_scf.m_set(sec_scf.m_get_quantity_definition(key), val)
 
         atom_forces = source.get('atom_forces', self.get_forces(source._frame))
         if atom_forces is not None:
@@ -1293,7 +1292,7 @@ class CP2KParser:
                     self.parse_system(atomic_coord)
                 else:
                     self.parse_system(n)
-                if sec_run.system[n] is not None:
+                if sec_run.system is not None and sec_run.system[n]:
                     sec_scc.system_ref = sec_run.system[n]
 
         single_point = quickstep.get('single_point')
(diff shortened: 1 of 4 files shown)


electronicparsers/cp2k/parser.py
```diff
@@ -1226,18 +1290,26 @@ def parse_calculations(calculations):
...
+                    self.parse_system(atomic_coord)
                 else:
-                    sec_system = self.parse_system(n)
-
-                if sec_system is not None:
-                    sec_scc.single_configuration_calculation_to_system_ref = sec_system
-
-            sec_scc.single_configuration_to_calculation_method_ref = sec_run.method[-1]
+                    self.parse_system(n)
+                if sec_run.system[n] is not None:
```
ladinesa: This would be problematic when no system is generated for step n.

electronicparsers/cp2k/parser.py
```diff
@@ -1057,11 +1051,116 @@ def get_xc_functionals(self):
...
+                name='CP2K', version=version[0] if isinstance(version, list) else version,
+                compilation_host=host[0] if isinstance(host, list) else host)
+            sec_program_information = sec_run.m_create(x_cp2k_section_program_information)
+            for key, val in cp2k_settings.items():
+                if key == 'svn_revision':
+                    try:
+                        val = int(val.strip('svn:'))
+                    except Exception:
+                        continue
+                setattr(sec_program_information, 'x_cp2k_%s' % key, val)
```
ladinesa: sec_program_information.m_set(sec_program_information.m_get_quantity_definition(f'x_cp2k_{key}'), val)

electronicparsers/cp2k/parser.py
```diff
@@ -1057,11 +1051,116 @@ def get_xc_functionals(self):
...
+
+        dbcsr_settings = self.settings.get('dbcsr', {})
+        if dbcsr_settings:
+            sec_dbcsr = sec_run.m_create(x_cp2k_section_dbcsr)
+            for key, val in dbcsr_settings.items():
+                setattr(sec_dbcsr, 'x_cp2k_%s' % key, val)
```
ladinesa: see previous comment

electronicparsers/cp2k/parser.py
```diff
@@ -1057,11 +1051,116 @@ def get_xc_functionals(self):
...
+            sec_startinformation = sec_run.m_create(x_cp2k_section_startinformation)
+            sec_endinformation = sec_run.m_create(x_cp2k_section_end_information)
+            section = sec_startinformation
+            for key, val in program_settings.items():
+                if key == 'id' and isinstance(val, list):
+                    sec_endinformation.x_cp2k_end_id = val[1]
+                    key, val = 'start_id', val[0]
+                section = sec_endinformation if key.startswith('end') else sec_startinformation
+                val = val[0] if isinstance(val, list) else val
+                setattr(section, 'x_cp2k_%s' % key, val)
```
ladinesa: see previous comment

electronicparsers/cp2k/parser.py
```diff
@@ -1057,11 +1051,116 @@ def get_xc_functionals(self):
...
+
+        global_settings = self.settings.get('global', {})
+        if global_settings:
+            sec_global_settings = sec_run.m_create(x_cp2k_section_global_settings)
+            for key, val in global_settings.items():
+                setattr(sec_global_settings, 'x_cp2k_%s' % key, val)
```
ladinesa: see previous comment

electronicparsers/cp2k/parser.py
```diff
@@ -1057,11 +1051,116 @@ def get_xc_functionals(self):
...
+                return name.replace('BASIS', 'BASIS_SET')
+            return name
+
+        def parse(name, data, section):
+            if isinstance(data, InpValue):
+                sec_def = resolve_definition(name)
+                if sec_def is not None:
+                    sub_section = section.m_create(sec_def.section_cls)
+                    for key, val in data.items():
+                        sec_name = '%s_%s' % (name, key)
```
ladinesa: parse(f'{name}_{key}', val, sub_section)

electronicparsers/cp2k/parser.py
JosePizarro3: done here and in the others

electronicparsers/cp2k/parser.py
JosePizarro3: done

electronicparsers/cp2k/parser.py
JosePizarro3: done. check now if the condition makes more sense.

JosePizarro3:
Thanks for the review. I am not sure what happened with the pipeline, some weird error messages when running pytest.

electronicparsers/cp2k/parser.py
ladinesa: it does not fix it since n_system is not strictly equal to n_calculation when you do not create a section system, this is the reason why I return sec_system.

ladinesa:
> Thanks for the review. I am not sure what happened with the pipeline, some weird error messages when running pytest.

probably with the latest changes I made, but my test passed, I rerun your pipeline, I hope it passes, otherwise I will have to rollback

electronicparsers/cp2k/parser.py
JosePizarro3: Do I leave out then the ref to system?

JosePizarro3:
> > Thanks for the review. I am not sure what happened with the pipeline, some weird error messages when running pytest.
> 
> probably with the latest changes I made, but my test passed, I rerun your pipeline, I hope it passes, otherwise I will have to rollback

It seems to be failing still with the same lengthy error on numpy.

electronicparsers/cp2k/parser.py
ladinesa: I suggest you return sec_system in parse_system if you want to put it as ref.

ladinesa:
> > > Thanks for the review. I am not sure what happened with the pipeline, some weird error messages when running pytest.
> > 
> > 
> > probably with the latest changes I made, but my test passed, I rerun your pipeline, I hope it passes, otherwise I will have to rollback
> 
> It seems to be failing still with the same lengthy error on numpy.

I just run the test and it seems to be coming from 

electronicparsers/cp2k/parser.py:1074: in parse_settings
    sec_program_information.m_set(sec_program_information.m_get_quantity_definition(f'x_cp2k_{key}'), val)

ValueError: Cannot find a proper definition for name x_cp2k_program_version

Please add this to cp2k metaifo def. I also removed the lengthy error in ams I will merge this to develop.

electronicparsers/cp2k/parser.py
JosePizarro3: Well, I am thinking that, if system and calcs lengths do not match, better not to ref. 

How do you or Joe handle these situations usually?

ladinesa: In md, it is actually more complicated since we should match the time step, so we do a map but here i think the solution is simple, you just have to return the created section in parse_system and ref it to the current section calc.

JosePizarro3:
> > > > Thanks for the review. I am not sure what happened with the pipeline, some weird error messages when running pytest.
> > > 
> > > 
> > > probably with the latest changes I made, but my test passed, I rerun your pipeline, I hope it passes, otherwise I will have to rollback
> > 
> > 
> > It seems to be failing still with the same lengthy error on numpy.
> 
> I just run the test and it seems to be coming from
> 
> electronicparsers/cp2k/parser.py:1074: in parse_settings sec_program_information.m_set(sec_program_information.m_get_quantity_definition(f'x_cp2k_{key}'), val)
> 
> ValueError: Cannot find a proper definition for name x_cp2k_program_version
> 
> Please add this to cp2k metaifo def. I also removed the lengthy error in ams I will merge this to develop.

This is surprising. I only moved this part of the code.

Anyways, I will instead define a JSON type to parse all this info.

ladinesa:
> This is surprising. I only moved this part of the code.

We were using setattr before hence no check is made.

JosePizarro3:
> > This is surprising. I only moved this part of the code.
> 
> We were using setattr before hence no check is made.

I went on deciding fixing the set attr for single point quantities, and left out GeomOpt and MD, as it is more involved and I am not sure what to leave out and what to keep.

tests/test_cp2kparser.py
```diff
@@ -33,16 +33,16 @@ def parser():
 
 def test_single_point(parser):
     archive = EntryArchive()
-    parser.parse('tests/data/cp2k/single_point/si_bulk8.out', archive, None)
+    # parser.parse('tests/data/cp2k/single_point/si_bulk8.out', archive, None)
+    parser.parse('/home/josepizarro/nomad/dependencies/parsers/electronic/tests/data/cp2k/single_point/si_bulk8.out', archive, None)
```
ladinesa: you probably forgot this

JosePizarro3: True, sorry.

electronicparsers/cp2k/parser.py
JosePizarro3: Check it out again, please.

ladinesa: looks good.

JosePizarro3:
Should I merge?

PR #149, opened 2023-08-18
Provide support for restart calcs in cp2k

ladinesa:


JosePizarro3 (commented):
I see there is some problem still finding the .restart file. As of now, the OutParser tries to read the restart filename and then find it in the calculation.

Can we make it more flexible using a smart combination of `self.out_parser.get('restart').get('filename', '')` and `get_files('*.restart', self.filepath, self.mainfile)`? Lines 1104-1108 of the parser.

ladinesa:
> I see there is some problem still finding the .restart file. As of now, the OutParser tries to read the restart filename and then find it in the calculation.
> 
> Can we make it more flexible using a smart combination of `self.out_parser.get('restart').get('filename', '')` and `get_files('*.restart', self.filepath, self.mainfile)`? Lines 1104-1108 of the parser.

I think if the restart file specified in the mainfile is not found, we should not try looking for alternatives. It is better not parsing anything than parsing the wrong one.

JosePizarro3:
> > I see there is some problem still finding the .restart file. As of now, the OutParser tries to read the restart filename and then find it in the calculation.
> > Can we make it more flexible using a smart combination of `self.out_parser.get('restart').get('filename', '')` and `get_files('*.restart', self.filepath, self.mainfile)`? Lines 1104-1108 of the parser.
> 
> I think if the restart file specified in the mainfile is not found, we should not try looking for alternatives. It is better not parsing anything than parsing the wrong one.

Yes, ok. I talked with Andrea, he will go and change the names of the files so it can be recognized. Our guess is that there is a script which changed the names to make them more readable, so it shouldn't be a problem to change it back.

JosePizarro3 (approved):

PR #150, opened 2023-08-18
148 handle changing symmetries in vasp

### LINKED ISSUE #148: Handle changing symmetries in VASP
During a geometry optimization or frequency calculation, the symmetry of the cell can change.
When symmetry-reduction is _active_, the number of reduced k-point can change by ionic update.
This causes errors when reshaping _eigenvalues_ extracted from the OUTCAR, which are contingent on the (original) no. irreducible k-points:

```
"root":{
  "event":string"Error reading eigenvalues"
  "proc":string"Entry"
  "process":string"process_entry"
  "process_worker_id":string"QrhFb34pRf6tYAkBJQqhpQ"
  "parser":string"parsers/vasp"
  "step":string"parsers/vasp"
  "logger":string"nomad.processing"
  "timestamp":string"2023-08-17 16:19.03"
  "level":string"ERROR"
}
```

The frequency calculation that brought this error to light: trash:///freqsafd_pbe.zip

ndaelman-hu:
Tackle changing symmetries (and k-point set) between ionic updates in newer VASP versions (about 6).
This solution packages `kpoints` into lists matching the no. `calculations`. The cases of single-point calculations and the older format are also handled (for `vasprun.xml`). No explicit information about the version is used.

Closes #148


ndaelman-hu:
@ladinesa Should I add the test referenced in the original issue?

electronicparsers/vasp/parser.py
```diff
@@ -514,14 +514,15 @@ def n_calculations(self):
...
-            self._kpoints_info = dict()
-            kpts_occs = self.parser.get('kpoints')
-            if kpts_occs is not None:
-                kpts_occs = np.reshape(kpts_occs, (len(kpts_occs) // 4, 4)).T
-                self._kpoints_info['points'] = kpts_occs[0:3].T
-                k_mults = kpts_occs[3].T
-                self._kpoints_info['multiplicities'] = k_mults
-                self._kpoints_info['weights'] = k_mults / np.sum(k_mults)
+            self._kpoints_info = []
+            for kpts_occs in self.parser.get('kpoints'):
```
ladinesa: put default, []

electronicparsers/vasp/parser.py
```diff
@@ -514,14 +514,15 @@ def n_calculations(self):
...
+                if kpts_occs is not None:
```
ladinesa: do you need this if?

electronicparsers/vasp/parser.py
```diff
@@ -514,14 +514,15 @@ def n_calculations(self):
...
+                    self._kpoints_info.append({})
```
ladinesa: probably define info = {} then append this later. Is there no info about the iter #? I am just worried that the kpoint does not match the step.  If there is none maybe put a check that len(kpoint_info) == len(calcs)

electronicparsers/vasp/parser.py
```diff
@@ -1032,39 +1033,51 @@ def n_calculations(self):
         'listgenerated': 'Line-path',
     }
 
+    def _find_kpoints(self) -> list[str]:
```
ladinesa: i do not see this used anywhere except in kpoints_info. I suggest not defining this.

electronicparsers/vasp/parser.py
```diff
@@ -1032,39 +1033,51 @@ def n_calculations(self):
...
+        stem = '/modeling[0]/calculation'
+        n_ionic_updates = len(self._get_key_values(stem).get('calculation', []))
```
ladinesa: where we can add iter no to info so we can really map to calc. may be we can do it also for outcar parser

electronicparsers/vasp/parser.py
```diff
@@ -1580,14 +1593,15 @@ def parse_energy(n_calc, n_scf=None):
 
         def parse_eigenvalues(n_calc):
             eigenvalues = self.parser.get_eigenvalues(n_calc)
+            kpoint_source = self.parser.kpoints_info[n_calc]
```
ladinesa: add error handling here if kpoints_info is None

electronicparsers/vasp/parser.py
```diff
@@ -1627,8 +1641,8 @@ def parse_eigenvalues(n_calc):
                 eigs = eigs * ureg.eV
                 sec_eigenvalues = sec_scc.m_create(BandEnergies)
                 sec_eigenvalues.kpoints = kpoints
-                sec_eigenvalues.kpoints_weights = self.parser.kpoints_info.get('weights', [])
-                sec_eigenvalues.kpoints_multiplicities = self.parser.kpoints_info.get('multiplicities', [])
+                sec_eigenvalues.kpoints_weights = kpoint_source.get('weights', [])
```
ladinesa: I suggest you put None as default

ladinesa:
Thanks for the fix. Please also add a test for this, maybe one outcar and one vasprun, you can also use this for the atom positions fix.

electronicparsers/vasp/parser.py
ndaelman-hu: Removed

electronicparsers/vasp/parser.py
ndaelman-hu: I've only seen 2 cases up to now: the k-points are only updated once, or they are updated at each ionic update.
I am storing the index now, and will consider using it for safety checks.

electronicparsers/vasp/parser.py
ndaelman-hu: Have moved it into `kpoints_info`

electronicparsers/vasp/parser.py
ndaelman-hu: I thought that an empty list is fine for _repeated_ quantities. Could we add this specification somewhere to the documentation?
Corrected both `weights` and `multiplicities`.

electronicparsers/vasp/parser.py
ndaelman-hu: Applied

electronicparsers/vasp/parser.py
ndaelman-hu: Could move it into `kpoints_info()`.
While I get the rationale behind reserving functions only for abstractions that are used more than once, I split it here for readability. This happens in other places of the VASP parser too. You let me know what you think.

electronicparsers/vasp/parser.py
ndaelman-hu: Btw, in `vasprun.xml`, it's rather the `eigenvalues` that are out of sync with the no. steps.

ndaelman-hu:
Something to consider in the future, when we clean the VASP parser.
The VASP mainfiles have very different structures. Just considering the case of the frequency analysis:
- `OUTCAR`: k-points and eigenvalues are repeated for each ionic update + k-points are repeated at the end
- `vasprun.xml`: k-points (list) are repeated at each ionic update + at the beginning (under `<incar>`). The (electronic) eigenvalues are only printed in the last calculations step (maybe also for under the `projected` tag)

When it comes to eigenvalues, we mostly just need the no. irreducible k-points (at that step) to restructure the eigenvalues.
Atm, we are extracting all the k-points in full, at each step, and this might be overkill. Full extraction is only needed at the last step.

electronicparsers/vasp/parser.py
ndaelman-hu: Pls, take a look at my comment below. Personally, I think we can go without any iteration index.
Or would you rather than a `list`, prefer a `dict` with a clearer index?

Instead, I'd rather have a way of moving `k-points` under `calculation`, but this isn't self-evident for `OUTCAR`.

PR #153, opened 2023-08-18
Fix atom positions 

### LINKED ISSUE #152: VASP 6 format: fix atom positions
The atom positions are not being properly extracted from the OUTCAR, leading to the error on line 197 in `nomad/normalizing/system.py` being triggered.

Same example as in #148

ndaelman-hu:
Update the atomic positions extraction to also account for the new VASP 6 format (where `0.0768` is written as `.0768`)

Closes #152


ladinesa (commented):
Can we perhaps define a regex expression for float in vasp, something like ref_f = r'[-+]?\d*\.*\d*(?:[Ee][-+]\d+)?' and use this throughout the code? It will make debugging a little easier. You can also not do it in this pr but please add a test for v6.

PR #154, opened 2023-08-22
TBStudio parser

### LINKED ISSUE #90: TBS parser
Hi @mohammadnakhaee , 

I created the skeleton of the parser in the branch `tbs_parser`. From here, you can define / populate your sections / quantities. Here I would like to stress again a couple of things:

- The difference between code-specific from general metainfo. For the code specific, you can always extend sections using `tbs/metainfo/tbs.py`. You can follow the conventions used in other parsers as an example, but the general idea is to define code-specific quantities with `x_codename_...`, hence in your case you could have `x_tbs_...`.
- Whenever you have any doubt, feel free to come to my office and we can check it together.
- Also drop docstrings whenever you can so we have a good maintainance of the parser on the long term :slightly_smiling_face: 

As you see, right now this branch is pointing to `1008-plugin-mechanism` as I branched out from it, but later we will have to rebase to `develop`. 

I will also include the MatchingParserInterface in the gitlab branch, so you understand where does it come from.

Once we finish an initial version, we pass it to Alvin for a final check.

JosePizarro3:


mohammadnakhaee:
Hi @JosePizarro3  I added the test for the parser. Please take a look if we can merge it.

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,498 @@
...
+        try:
+            x = np.float64(r[0])
+            y = np.float64(r[1])
+            z = np.float64(r[2])
+            xyz_coords.append([x, y, z])
+        except:
+            break
+    model['xyz_coords'] = xyz_coords
+
+    tb_l = parse_int(tbm['vars']['TBl[0]'], 0)
```
JosePizarro3: Did you try `tb_l = np.int64(tbm['vars'].get('TBl[0]', '0'))`? Here and every time you used `parse_int` or `parse_float`

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,498 @@
...
+    def parse(self, filepath, archive, logger):
+        tb_archive = archive
+        self.filepath = os.path.abspath(filepath)
+        self.archive = tb_archive
+        self.maindir = os.path.dirname(self.filepath)
+        self.logger = logger if logger is not None else logging
+
+        sec_run = self.archive.m_create(Run)
+        sec_run.program = Program(name="TBStudio")
+        self.tb_model = load_tbm(filepath)
```
JosePizarro3: I think there is a way of retrieving binary files keys and values. @ladinesa can you confirm this or should Mohammad be using this `load_tbm` as it is now?

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,498 @@
...
+def parse_float(string, default):
+    val = default
+    try:
+        val = np.float64(string)
+    except:
+        pass
+    return val
+
+
+def load_tbm(file):
```
JosePizarro3: I think indeed there is a better way of defining the parser. Can you please check `dependencies/parsers/workflow/fhivibes/parser.py`? In there, there is a class called `XarrayParser(FileParser)` which can be used for you here.

So I suggest:

1. Define here `TbmParser(FileParser)` and initialize in `TBStudioParser.__init__`. You can call there `self.parser = TbmParser()`. 
2. At the same time, define a new function in `TBStudioParser` call `init_parameters(self)` which:
```
    def init_parser(self):
        self.parser.mainfile = self.filepath
        self.parser.logger = self.logger
```
3. Now, `TbmParser` has to have several functions extracting information (similarly to `fhivibes/parser.py:XarrayParser`.

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,498 @@
...
+    def parse_workflow(self):
+        sec_workflow = self.archive.m_create(Workflow)
+        workflow = SinglePoint()
+        sec_workflow.type = 'single_point'
+        workflow.name = "Tight Binding Model"
+        self.archive.workflow2 = workflow
+
+    def get_mainfile_keys(self, **kwargs):
+        filepath = kwargs.get('filename')
+        tb_model = load_tbm(filepath)
```
JosePizarro3: One of the advantages of packing load_tbm in a class, is that in this case you are reparsing the whole model. This is not very optimized.

The suggestion here is to create a @property in `TbmParser` which can retrieve a tuple with the DFTNomadEntryID and the DFTNomadUploadID, and then you use it here. 

Still, you would need to initialize the parser with `self.parser.mainfile = filepath`. Is this correct, @ladinesa or this should be more careful taking into consideration with decoder and buffer?

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,498 @@
...
+
+    # Load workflow if does exist
+    model['DFTNomadEntryID'] = None
+    model['DFTNomadUploadID'] = None
+    if 'type' in tbm['DFTSource'] and tbm['DFTSource']['type'].lower() == 'nomad':
+        model['DFTNomadUploadID'] = tbm['DFTSource']['source']['upload_id']
+        model['DFTNomadEntryID'] = tbm['DFTSource']['source']['entry_id']
+
+    # Load lattice vectors
+    model['a'] = [parse_float(tbm['vars']['a[0]'], 0), parse_float(tbm['vars']['a[1]'], 0), parse_float(tbm['vars']['a[2]'], 0)]
```
JosePizarro3: I think you can automatize this a bit with a for loop.

```
model = {}
for axis in ['a', 'b', 'c']:
   model[axis] = [np.float64(tbm['vars'].get(f'a[{i}]', '0')) for i in range(2)]
```

Furthermore, you can define (for readability), `tbm_variables = tbm.get('vars')` and make sure they exist, i.e., before this for loop have an if that returns nothing:
```
if tbm.get('vars') is None: 
   self.logger.warning('Could not find the .tbm file vars parameters.')
   return

model = {}
for axis in ['a', 'b', 'c']:
   model[axis] = [np.float64(tbm['vars'].get(f'a[{i}]', '0')) for i in range(2)]
```

This is a typical procedure to ensure that the i/o are read and there is not errors popping up when parsing (essentially, error handling).

electronicparsers/tbstudio/parser.py
JosePizarro3: I think if you use the parser similar to `XarrayParser` in fhivibes, you will get rid off this uncomfy situations 🙂

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,498 @@
...
+            xyz_coords.append([x, y, z])
+        except:
+            break
+    model['xyz_coords'] = xyz_coords
+
+    tb_l = parse_int(tbm['vars']['TBl[0]'], 0)
+    tb_m = parse_int(tbm['vars']['TBm[0]'], 0)
+    tb_n = parse_int(tbm['vars']['TBn[0]'], 0)
+
+    model['neighbor_unit_cells'] = [tb_l, tb_m, tb_n]
```
JosePizarro3: You can further automatize this (similar to the for loop before).

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,498 @@
...
+    model['final_os'] = {}
+    tbAtom = ''
+    shell = ''
+    for state, row in zip(_osr, _os):
+        if state[0] and state[1] and state[2]:
+            if row[0] == '' and row[1] == '' and row[2] == '':
+                break
+            else:
+                os_name = row[0]
+                orbital_info = re.search("^(.*) \((.*)\)$", os_name)
```
JosePizarro3: I am surprised you have to re.search. But again, this comes back to use a class which inherits from FileParser. Then you can use the Quantity functionality for regex in there :)

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,498 @@
...
+    k_length = []
+    for row in _k_points:
+        frac_k_points.append([row[0], row[1], row[2]])
+        k_points.append([row[3], row[4], row[5]])
+        k_length.append(row[6])
+    model['frac_k_points'] = frac_k_points
+    model['k_points'] = k_points
+    model['k_length'] = k_length
+
+    model['bandSections'] = tbm['variables']['bandSections']
```
JosePizarro3: You can automatize this as well with a mapping:
```
model_variables_map = {
   'bandSections': 'bandSections',
   'DFTEigVal': 'dft_bands',
   'iTBEigVal': 'initial_tb_bands',
   'fTBEigVal': 'final_tb_bands',
   'ChemP': 'fermi_level'
}

tbm_variables = tbm.get('variables', {})
for keys, values in model_variables_map.items():
   model[values] = tbm_variables.get(keys)
```

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,498 @@
...
+    model['initial_tb_bands'] = tbm['variables']['iTBEigVal']
+    model['final_tb_bands'] = tbm['variables']['fTBEigVal']
+    model['fermi_level'] = tbm['variables']['ChemP']
+
+    tb_unit_cells = tbm['lists']['EssentialUnitcellList']
+    model['initial_hamiltonian_matrix'] = {}
+    model['final_hamiltonian_matrix'] = {}
+    model['initial_overlap_matrix'] = {}
+    model['final_overlap_matrix'] = {}
+    for index, unit_cell in enumerate(tb_unit_cells):
```
JosePizarro3: Here you can further define another loop and map.

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,498 @@
...
+        if len(tbm['variables']['Hi']) > index:
+            model['initial_hamiltonian_matrix'][unit_cell] = tbm['variables']['Hi'][index]
+        if len(tbm['variables']['Hi']) > index:
```
JosePizarro3: typo: I think the key is 'Hf'

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,498 @@
...
+            model['final_hamiltonian_matrix'][unit_cell] = tbm['variables']['Hf'][index]
+        if len(tbm['variables']['Si']) > index:
+            model['initial_overlap_matrix'][unit_cell] = tbm['variables']['Si'][index]
+        if len(tbm['variables']['Sf']) > index:
+            model['final_overlap_matrix'][unit_cell] = tbm['variables']['Sf'][index]
+
+    model['initial_soc_matrix'] = {}
+    model['final_soc_matrix'] = {}
+
+    if len(tbm['variables']['SOCi']) > 0:
```
JosePizarro3: For loop please

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,498 @@
...
+
+    def __init__(self):
+        self._calculation_type = 'tight binding'
+
+    def parse_system(self):
+        """Populates run.system with the input structural parameters.
+        """
+        sec_run = self.archive.run[-1]
+        sec_system = sec_run.m_create(System)
+        # Here the key is to populate:
```
JosePizarro3: Delete comments please

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,498 @@
...
+
+        for n1, n2 in band_segments_points:
+            sec_k_band_segment = sec_k_band.m_create(BandEnergies)
+            sec_k_band_segment.kpoints = frac_k_points[n1: n2 + 1]
+            sec_k_band_segment.energies = (np.array([tb_bands[n1: n2 + 1]]) + self.tb_model['fermi_level']) * ureg.eV
+
+        sec_scc.band_structure_electronic.append(sec_k_band)
+
+    def parse_workflow(self):
+        sec_workflow = self.archive.m_create(Workflow)
```
JosePizarro3: You can deprecate sec_workflow and not use it. Simply do `workflow` and `self.archive.workflow2 = workflow`.

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,498 @@
...
+        workflow = SinglePoint()
+        sec_workflow.type = 'single_point'
+        workflow.name = "Tight Binding Model"
```
JosePizarro3: No need to write a name. The typical naming of the workflow is `SinglePoint`, and we should not change this. Then it is the Method what establishes the type of `SinglePoint`.

So, in short, you can delete this line.

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,498 @@
...
+        sec_run.program = Program(name="TBStudio")
+        self.tb_model = load_tbm(filepath)
+        dft_nomad_entry_id = self.tb_model['DFTNomadEntryID']
+        dft_nomad_upload_id = self.tb_model['DFTNomadUploadID']
+
+        self.parse_system()
+        self.parse_method()
+        self.parse_scc()
+
+        self.parse_workflow()
```
JosePizarro3: No need to define a specific function here. Simply delete that method and do:
```
workflow = SinglePoint()
self.archive.workflow2 = workflow
```

electronicparsers/utils/utils.py
```diff
@@ -161,6 +160,56 @@ def parse_gw_workflow(self, gw_archive: EntryArchive, gw_workflow_archive: Entry
...
+            workflow.m_add_sub_section(
+                TB.inputs, Link(name='Structure', section=input_structure))
+        if tb_calculation:
+            workflow.m_add_sub_section(
+                TB.outputs, Link(name='TB Model', section=tb_calculation))
+
+        # First Principles Calculation task
+        if self.archive.workflow2:
+            first_principles_task = TaskReference(task=first_principles_calculation_archive.workflow2)
+            first_principles_task.name = 'First-Principles Calculation'
```
JosePizarro3: No need to add Calculation, you can keep it to `'First principles'` or `'Ab initio'`.

electronicparsers/utils/utils.py
```diff
@@ -161,6 +160,56 @@ def parse_gw_workflow(self, gw_archive: EntryArchive, gw_workflow_archive: Entry
...
+            if input_structure:
+                first_principles_task.inputs = [Link(name='Structure', section=input_structure)]
```
JosePizarro3: Please, add specifically `'Input structure'`.

electronicparsers/utils/utils.py
```diff
@@ -161,6 +160,56 @@ def parse_gw_workflow(self, gw_archive: EntryArchive, gw_workflow_archive: Entry
...
+            if first_principles_calculation:
+                first_principles_task.outputs = [Link(name='First Principles Calculation', section=first_principles_calculation)]
```
JosePizarro3: `'Output first principles calculation'`.

electronicparsers/utils/utils.py
```diff
@@ -161,6 +160,56 @@ def parse_gw_workflow(self, gw_archive: EntryArchive, gw_workflow_archive: Entry
...
+            workflow.m_add_sub_section(TB.tasks, first_principles_task)
+
+        # TB task
+        if tb_archive.workflow2:
+            tb_task = TaskReference(task=tb_archive.workflow2)
+            tb_task.name = 'Tight-Binding'
+            if first_principles_calculation:
+                tb_task.inputs = [Link(name='First-Principles Calculation', section=first_principles_calculation)]
```
JosePizarro3: `'Input first principles calculation'`

electronicparsers/utils/utils.py
```diff
@@ -161,6 +160,56 @@ def parse_gw_workflow(self, gw_archive: EntryArchive, gw_workflow_archive: Entry
...
+            if tb_calculation:
+                tb_task.outputs = [Link(name='TB Model', section=tb_calculation)]
```
JosePizarro3: `'Output TB calculation'`

electronicparsers/utils/utils.py
```diff
@@ -161,6 +160,56 @@ def parse_gw_workflow(self, gw_archive: EntryArchive, gw_workflow_archive: Entry
...
+            if input_structure:
+                first_principles_task.inputs = [Link(name='Structure', section=input_structure)]
+            if first_principles_calculation:
+                first_principles_task.outputs = [Link(name='First Principles Calculation', section=first_principles_calculation)]
+            workflow.m_add_sub_section(TB.tasks, first_principles_task)
+
+        # TB task
+        if tb_archive.workflow2:
+            tb_task = TaskReference(task=tb_archive.workflow2)
+            tb_task.name = 'Tight-Binding'
```
JosePizarro3: `tb_task.name = 'TB'`.

electronicparsers/utils/utils.py
```diff
@@ -161,6 +160,56 @@ def parse_gw_workflow(self, gw_archive: EntryArchive, gw_workflow_archive: Entry
...
+        # Inputs and Outputs
+        input_structure = extract_section(first_principles_calculation_archive, 'run/system')
+        first_principles_calculation = extract_section(first_principles_calculation_archive, 'run/calculation')
+        tb_calculation = extract_section(tb_archive, 'run/calculation')
+        if input_structure:
+            workflow.m_add_sub_section(
+                TB.inputs, Link(name='Structure', section=input_structure))
+        if tb_calculation:
+            workflow.m_add_sub_section(
+                TB.outputs, Link(name='TB Model', section=tb_calculation))
```
JosePizarro3: The name has to be consistent. `name='Output TB calculation'`.

electronicparsers/utils/utils.py
```diff
@@ -161,6 +160,56 @@ def parse_gw_workflow(self, gw_archive: EntryArchive, gw_workflow_archive: Entry
...
+        method_tb = extract_section(tb_archive, 'run/method/tb')
+        workflow.method.tb_method_ref = method_tb
+
+        # Inputs and Outputs
+        input_structure = extract_section(first_principles_calculation_archive, 'run/system')
+        first_principles_calculation = extract_section(first_principles_calculation_archive, 'run/calculation')
+        tb_calculation = extract_section(tb_archive, 'run/calculation')
+        if input_structure:
+            workflow.m_add_sub_section(
+                TB.inputs, Link(name='Structure', section=input_structure))
```
JosePizarro3: `name='Input structure'`

electronicparsers/utils/utils.py
```diff
@@ -161,6 +160,56 @@ def parse_gw_workflow(self, gw_archive: EntryArchive, gw_workflow_archive: Entry
...
+
+        Args:
+            tb_archive (EntryArchive): the Tight-Binding archive
+            first_principles_calculation_archive (EntryArchive): the first-principles-calculation archive
+            tb_workflow_archive (EntryArchive): the Tight-Binding workflow archive
+        """
+        self.run_workflow_archive(tb_workflow_archive)
+        tb_workflow_archive.run[-1].m_add_sub_section(Run.system, first_principles_calculation_archive.run[-1].system[-1])
+        workflow = TB(method=TBMethod())
+        workflow.name = 'TB Fitting'
```
JosePizarro3: Simply keep it as `'TB'`. When talking about workflows, the name is a simple label that specifies what has been done. If we start putting sub-labels, we might be complicating queries later on.

JosePizarro3 (commented):
Very nice, just some more detail work and the parser will be ready.

I've only made an initial review, focusing on the parsing side. Once you have the new changes under control, I will turn to the specific parsing and metainfo definitions (everything that is inside parse_system, parse_method, parse_scc + the metadata in nomad).

Let me know if something is unclear.

electronicparsers/tbstudio/parser.py
mohammadnakhaee: The tbm file is not binary anymore. It is a JSON file in the new version

JosePizarro3: In that case, it should be simply be 
```
try:
   data = json.load(open(self.filepath))
except Exception:
   self.logger.error('Error opening the json output mainfile.')
   data = None
   return
self.tb_model = data
```

And you need to `import json`.

electronicparsers/tbstudio/parser.py
mohammadnakhaee: json.load is used directly in the parser class

electronicparsers/tbstudio/parser.py
mohammadnakhaee: I applied the first loop. I did not use a loop for the axis name since the middle dictionary model has been removed now

mohammadnakhaee:
Hi @JosePizarro3 Please take a look on the last changes.

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+        sec_run = self.archive.run[-1]
+        sec_system = sec_run.m_create(System)
+        sec_atoms = sec_system.m_create(Atoms)
+
+        # Load lattice vectors
+        if self.tbm.get('vars') is None:
+            self.logger.warning('Could not find the .tbm file vars parameters.')
+            return
+
+        self.a = [np.float64(self.tbm['vars'].get(f'a[{i}]', '0')) for i in range(3)]
```
JosePizarro3, line 60: I think you can get rid of the self tag, and even further automatize this:
```
lattice_vectors = []
for axis in ['a', 'b', 'c']:
   lattice_vectors.append([np.float64(self.tbm['vars'].get(f'{axis}[{i}]', '0')) for i in range(3)])
sec_atoms.lattice_vectors = lattice_vectors * ureg.angstrom
```

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+        self.b = [np.float64(self.tbm['vars'].get(f'b[{i}]', '0')) for i in range(3)]
+        self.c = [np.float64(self.tbm['vars'].get(f'c[{i}]', '0')) for i in range(3)]
+
+        sec_atoms.lattice_vectors = [self.a, self.b, self.c] * ureg.angstrom
+
+        xyz_coords = []
+        for r in self.tbm['grids']['XYZ_Coords']['value']:
+            try:
```
JosePizarro3, line 71: Here, you are checking that r[i] can be transformed into a np.float64, right? I think the try is correct, but maybe you can add a comment to specify this :-)

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+                y = np.float64(r[1])
+                z = np.float64(r[2])
+                xyz_coords.append([x, y, z])
+            except:
+                break
+
+        sec_atoms.positions = xyz_coords * ureg.angstrom
+
+        kinds = []
+        for r in self.tbm['grids']['KABC_Coords']['value']:
```
JosePizarro3: What about using a .get here?
```
kinds = []
for r in self.tbm['grids']['KABC_Coords'].get('value'):
   if r is not None:
      kinds.append(np.int64(r[0]))
   else:
      break
```

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+            except:
+                break
+
+        sec_atoms.species = kinds
+
+        tb_l = np.int64(self.tbm['vars'].get('TBl[0]', '0'))
+        tb_m = np.int64(self.tbm['vars'].get('TBm[0]', '0'))
+        tb_n = np.int64(self.tbm['vars'].get('TBn[0]', '0'))
+
+        pbc = [bool(dim != 0) for dim in [tb_l, tb_m, tb_n]]
```
JosePizarro3, line 88: I think you could use `lattice_vectors` instead of having to calculate `[tb_l, tb_m, tb_n]`, right?

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+
+        kinds = []
+        for r in self.tbm['grids']['KABC_Coords']['value']:
+            try:
+                k = np.int64(r[0])
+                kinds.append(k)
+            except:
+                break
+
+        sec_atoms.species = kinds
```
JosePizarro3: Is this `species` or `labels`? The difference is that species only refers to non-repeated atoms. For example, in graphene `species = 'C'` while `labels = ['C', 'C']`, so that labels refer to the two atoms in the unit cell.

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+
+        pbc = [bool(dim != 0) for dim in [tb_l, tb_m, tb_n]]
+        sec_atoms.periodic = pbc
+
+    def parse_method(self):
+        """Populates run.method with the input methodological parameters.
+        """
+        sec_run = self.archive.run[-1]
+
+        sec_run.program = Program(name='TBStudio', version=self.tbm['ReleaseVersion'])
```
JosePizarro3: Error handling:
```
sec_run.program = Program(name='TBStudio', version=self.tbm.get('ReleaseVersion', ''))
```

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+
+        sec_method = sec_run.m_create(Method)
+        sec_tb = sec_method.m_create(TB)
+        sec_sk = sec_tb.m_create(SlaterKoster)
```
JosePizarro3, line 98: Can you check if you can directly create:
```
sec_method = sec_run.m_create(Method)
sec_sk = sec_method.m_create(TB).m_create(SlaterKoster)
```
?

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+                indices = re.findall(r'-?\d+', atom2['cell'])
+                center2.cell_index = [int(index) for index in indices]
+                for sk_label, sk_integral in s_sk.items():
+                    setattr(sec_overlaps, sk_label, sk_integral)
+
+    def parse_scc(self):
+        """Populates run.calculation with the output of the calculation.
+        """
+        sec_run = self.archive.run[-1]
+        self.archive.run[-1].m_create(Calculation)
```
JosePizarro3: You can:
```
sec_scc = sec_run.m_create(Calculation)
```

And delete the line here where you do `sec_scc = sec_run.calculation[-1]` (see next comment).

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+            k_points.append([row[3], row[4], row[5]])
+            k_length.append(row[6])
+
+        tb_bands = self.tbm['variables']['fTBEigVal']
+        band_segments_points = self.tbm['variables']['bandSections']['index']
+        if band_segments_points is None or len(tb_bands) < 1 or len(frac_k_points) < 1:
+            return
+
+        fermi_level_joules = self.tbm['variables']['ChemP'] * ureg.eV
+        sec_scc = sec_run.calculation[-1]
```
JosePizarro3: You can delete this when doing what I said in the comment before.

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+            frac_k_points.append([row[0], row[1], row[2]])
+            k_points.append([row[3], row[4], row[5]])
+            k_length.append(row[6])
+
+        tb_bands = self.tbm['variables']['fTBEigVal']
+        band_segments_points = self.tbm['variables']['bandSections']['index']
+        if band_segments_points is None or len(tb_bands) < 1 or len(frac_k_points) < 1:
+            return
+
+        fermi_level_joules = self.tbm['variables']['ChemP'] * ureg.eV
```
JosePizarro3: Error handling (and no need to name 'joules'):
```
fermi_level = self.tbm['variables'].get('ChemP')
sec_energy = sec_scc.m_create(Energy, Calculation.energy)
sec_energy.fermi = fermi_level * ureg.eV if fermi_level else None
```

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+        sec_scc = sec_run.calculation[-1]
+        sec_energy = sec_scc.m_create(Energy, Calculation.energy)
+        sec_energy.fermi = fermi_level_joules
+
+        sec_k_band = BandStructure()
```
JosePizarro3, line 290: I think you should check if the fermi_level was extracted in order to generate the `BandStructure`. Something like:
```
if not fermi_level:
   self.logger.warning('Could not extract the Fermi level, so that the BandStructure is not resolved')
   return
sec_k_band = BandStructure()
...(the rest of your code)
```

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+        sec_k_band.energy_fermi = fermi_level_joules
+
+        pi = np.arccos(-1.0)
+        vol = np.dot(self.a, np.cross(self.b, self.c))
+        astar = 2 * pi * np.cross(self.b, self.c) / vol * 10**9
+        bstar = 2 * pi * np.cross(self.c, self.a) / vol * 10**9
+        cstar = 2 * pi * np.cross(self.a, self.b) / vol * 10**9
+        sec_k_band.reciprocal_cell = [astar, bstar, cstar]
```
JosePizarro3: You don't need this. The normalizer will take care of it by reading `System`, so please, delete these lines.

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+
+        fermi_level_joules = self.tbm['variables']['ChemP'] * ureg.eV
+        sec_scc = sec_run.calculation[-1]
+        sec_energy = sec_scc.m_create(Energy, Calculation.energy)
+        sec_energy.fermi = fermi_level_joules
+
+        sec_k_band = BandStructure()
+        sec_k_band.energy_fermi = fermi_level_joules
+
+        pi = np.arccos(-1.0)
```
JosePizarro3: You can delete this too, but otherwise, pi is defined in numpy, as `np.pi`.

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+        bstar = 2 * pi * np.cross(self.c, self.a) / vol * 10**9
+        cstar = 2 * pi * np.cross(self.a, self.b) / vol * 10**9
+        sec_k_band.reciprocal_cell = [astar, bstar, cstar]
+
+        for n1, n2 in band_segments_points:
+            sec_k_band_segment = sec_k_band.m_create(BandEnergies)
+            sec_k_band_segment.kpoints = frac_k_points[n1: n2 + 1]
+            sec_k_band_segment.energies = (np.array([tb_bands[n1: n2 + 1]]) + self.tbm['variables']['ChemP']) * ureg.eV
+
+        sec_scc.band_structure_electronic.append(sec_k_band)
```
JosePizarro3, line 298: One tricky detail here. When appending repeated sub-sections to a parent section you have to use `m_add_sub_section` instead.

It will look like:
```
sec_scc.m_add_sub_section(Calculation.band_structure_electronic, sec_k_band)
```

This is true for any other appending of sections in NOMAD.

electronicparsers/tbstudio/parser.py
JosePizarro3: Please, delete these old lines. Simply add:
```
workflow = SinglePoint()
self.archive.workflow2 = workflow
```

Everything else is not used anymore. And no need to change the name.

electronicparsers/tbstudio/parser.py
JosePizarro3: Please, delete.

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+        sec_workflow = self.archive.m_create(Workflow)
+        workflow = SinglePoint()
+        sec_workflow.type = 'single_point'
+        workflow.name = "Tight Binding Model"
+        self.archive.workflow2 = workflow
+
+    def get_mainfile_keys(self, **kwargs):
+        filepath = kwargs.get('filename')
+
+        f = open(filepath)
```
JosePizarro3, line 303: You don't need to open filepath. Simply do `tbm = json.load(filepath)`

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+        tbm = json.load(f)
+        f.close()
+
+        dft_nomad_entry_id = None
```
JosePizarro3, line 307: You can compact these ifs into:
```
dft_source = tbm.get('DFTSource', {})
if dft_source.get('type') and dft_source.get('type').lower() == 'nomad':
   dft_nomad_entry_id = dft_source['source']['entry_id']
   dft_nomad_upload_id = dft_source['source']['upload_id']
   if dft_nomad_entry_id and dft_nomad_upload_id:
      return ['TB_workflow']
return True
```

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+            dft_nomad_entry_id = tbm['DFTSource']['source']['entry_id']
+            dft_nomad_upload_id = tbm['DFTSource']['source']['upload_id']
+
+        if dft_nomad_entry_id is not None and dft_nomad_entry_id != '' and dft_nomad_upload_id is not None and dft_nomad_upload_id != '':
+            return ['TB_workflow']
+        else:
+            return True
+
+    def parse(self, filepath, archive, logger):
+        tb_archive = archive
```
JosePizarro3: No need to write this line.

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+        self.filepath = os.path.abspath(filepath)
+        self.archive = tb_archive
```
JosePizarro3: Change by `self.archive = archive`.

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+            dft_nomad_upload_id = tbm['DFTSource']['source']['upload_id']
+
+        if dft_nomad_entry_id is not None and dft_nomad_entry_id != '' and dft_nomad_upload_id is not None and dft_nomad_upload_id != '':
+            return ['TB_workflow']
+        else:
+            return True
+
+    def parse(self, filepath, archive, logger):
+        tb_archive = archive
+        self.filepath = os.path.abspath(filepath)
```
JosePizarro3, line 319: No need, do `self.filepath = filepath`. Did you do this because of debugging?

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+        self.archive = tb_archive
+        self.maindir = os.path.dirname(self.filepath)
+        self.logger = logger if logger is not None else logging
+
+        sec_run = self.archive.m_create(Run)
```
JosePizarro3, line 324: Before defining Run, do a check on loading the json data:
```
        try:
            tbm = json.load(open(self.filepath))
        except Exception:
            self.logger.error('Error opening json output file.')
            tbm = None
            return
        self.tbm = tbm
```

No need to open the filepath.

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+        sec_run.program = Program(name="TBStudio")
```
JosePizarro3: Furthermore, you were written the Program information in `parse_method`. Please, move the program definition here (deleting the corresponding lines in `parse_method`) doing:
```
sec_run.program = Program(name='TBStudio', version=self.tbm.get('ReleaseVersion', ''))
```

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+        if 'type' in self.tbm['DFTSource'] and self.tbm['DFTSource']['type'].lower() == 'nomad':
+            dft_nomad_entry_id = self.tbm['DFTSource']['source']['entry_id']
+            dft_nomad_upload_id = self.tbm['DFTSource']['source']['upload_id']
+
+        self.parse_system()
+        self.parse_method()
+        self.parse_scc()
+
+        self.parse_workflow()
+        if dft_nomad_entry_id is not None and dft_nomad_entry_id != '' and dft_nomad_upload_id is not None and dft_nomad_upload_id != '':
```
JosePizarro3, line 350: Please, see my comment in `get_mainfile_keys` and do the same here.

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+            first_principles_calculation_archive = None
+            try:
+                first_principles_calculation_archive = archive.m_context.load_archive(dft_nomad_entry_id, dft_nomad_upload_id, None)
+            except:
+                pass
+            if first_principles_calculation_archive and self._child_archives:
+                tb_workflow_archive = self._child_archives.get('TB_workflow')
+                self.parse_tb_workflow(tb_archive, first_principles_calculation_archive, tb_workflow_archive)
```
JosePizarro3: No need to call tb_archive, but rather pass `archive` or `self.archive`

electronicparsers/tbstudio/parser.py
```diff
@@ -0,0 +1,376 @@
...
+        self.parse_method()
+        self.parse_scc()
+
+        self.parse_workflow()
+        if dft_nomad_entry_id is not None and dft_nomad_entry_id != '' and dft_nomad_upload_id is not None and dft_nomad_upload_id != '':
+            first_principles_calculation_archive = None
+            try:
+                first_principles_calculation_archive = archive.m_context.load_archive(dft_nomad_entry_id, dft_nomad_upload_id, None)
+            except:
+                pass
```
JosePizarro3: I think you dont need this try. Do instead:
```
first_principles_calculation_archive = archive.m_context.load_archive(dft_nomad_entry_id, dft_nomad_upload_id, None)
if first_principles_calculation_archive:
   tb_workflow_archive = self._child_archives.get('TB_workflow')
   if tb_workflow_archive:
      self.parse_tb_workflow(...)
```

If you check `self._child_archives` you will probably get a pass as `get_mainfile_keys` is returning true.

electronicparsers/utils/utils.py
```diff
@@ -161,6 +160,56 @@ def parse_gw_workflow(self, gw_archive: EntryArchive, gw_workflow_archive: Entry
...
+            first_principles_calculation_archive (EntryArchive): the first-principles-calculation archive
+            tb_workflow_archive (EntryArchive): the Tight-Binding workflow archive
+        """
+        self.run_workflow_archive(tb_workflow_archive)
+        tb_workflow_archive.run[-1].m_add_sub_section(Run.system, first_principles_calculation_archive.run[-1].system[-1])
+        workflow = TB(method=TBMethod())
+        workflow.name = 'TB'
+
+        # Method
+        method_tb = extract_section(tb_archive, 'run/method/tb')
```
JosePizarro3, line 175: Is not interesting to also extract the First principles method section so it is referenced to the specific DFT or GW methodology followed?

electronicparsers/utils/utils.py
JosePizarro3: The reason of naming a task with 'Calculation' might be confusing. Even thought we all know that calculation is what we do, in NOMAD we further define 'calculation' as containing the outputs of a given 'method' applied to a specific 'material'. Here we are talking about the global task (containing system, method, calculation), hence I think it makes more sense to simply call it `'First Principles'`

electronicparsers/utils/utils.py
```diff
@@ -161,6 +160,56 @@ def parse_gw_workflow(self, gw_archive: EntryArchive, gw_workflow_archive: Entry
...
+            if first_principles_calculation:
+                first_principles_task.outputs = [Link(name='Output First Principles Calculation', section=first_principles_calculation)]
+            workflow.m_add_sub_section(TB.tasks, first_principles_task)
+
+        # TB task
+        if tb_archive.workflow2:
+            tb_task = TaskReference(task=tb_archive.workflow2)
+            tb_task.name = 'TB'
+            if first_principles_calculation:
+                tb_task.inputs = [Link(name='Input First-Principles Calculation', section=first_principles_calculation)]
```
JosePizarro3: Please, delete hyphon: `'Input First Principles Calculation'`.

electronicparsers/utils/utils.py
```diff
@@ -161,6 +160,56 @@ def parse_gw_workflow(self, gw_archive: EntryArchive, gw_workflow_archive: Entry
...
+        # Inputs and Outputs
+        input_structure = extract_section(first_principles_calculation_archive, 'run/system')
+        first_principles_calculation = extract_section(first_principles_calculation_archive, 'run/calculation')
+        tb_calculation = extract_section(tb_archive, 'run/calculation')
+        if input_structure:
+            workflow.m_add_sub_section(
+                TB.inputs, Link(name='Input Structure', section=input_structure))
+        if tb_calculation:
+            workflow.m_add_sub_section(
+                TB.outputs, Link(name='Output TB Model', section=tb_calculation))
```
JosePizarro3, line 187: Naming should be consistent with inputs/outputs. You called 'Output TB Model' and 'Output TB Calculation'.

tests/test_tbstudioparser.py
```diff
@@ -0,0 +1,98 @@
...
+
+    assert len(sec_run.system) == 1
+    sec_system = sec_run.system[-1]
+    assert (sec_system.atoms.species == [6, 6]).all()
+
+    a = [1.23435, 2.14452, 0.0]
+    b = [1.23435, -2.14452, 0.0]
+    c = [0.0, 0.0, 20.0]
+    positions = [[0.00414, -0.004863, 10.0], [1.238611, -0.72006, 10.0]]
+    assert sec_system.atoms.lattice_vectors.magnitude * 10 ** 10 == approx(np.array([a, b, c]))
```
JosePizarro3: you can pass to other units using `to('<units>')`. Here:
```
assert sec_system.atoms.lattice_vectors.to('angstrom').magnitude == approx(np.array([a, b, c]))
```

tests/test_tbstudioparser.py
```diff
@@ -0,0 +1,98 @@
...
+    assert sec_system.atoms.positions.magnitude * 10 ** 10 == approx(np.array(positions))
```
JosePizarro3: Same here.

JosePizarro3 (commented):
Hi,

Here you have a new round of comments. Mainly coding recommendations, a few checks, and some stuff that would make your life easier if you need to maintain this parser 🙂 

Feel free to ask back. When you go resolving comments, feel free to drop a message like 'Done', or some more extended explanation, and resolve it yourself.

Thanks a lot for the hard work, the parser looks very nice! 👍🏻  And after these changes, I can review the nomad-gitlab side and we could merge soon.

electronicparsers/tbstudio/parser.py
mohammadnakhaee: I need the self.a, self.b, and self.c somewhere else. That is why I did not loop over them

electronicparsers/tbstudio/parser.py
mohammadnakhaee: I combined it with the next loop and commented on both of them since they have the same strategy

electronicparsers/tbstudio/parser.py
mohammadnakhaee: It is a dictionary they have the same behaviour unless we want to provide a default value.

electronicparsers/tbstudio/parser.py
mohammadnakhaee: These parameters could be customized in tbstudio even we have a 3d structure we can reduce it to a 2d tb model

electronicparsers/tbstudio/parser.py
mohammadnakhaee: Good point. I did not know about the difference. I have the atomic numbers and now I convert them to atomic labels. I changed the tests as well. I guess the species will be calculated automatically, right?

electronicparsers/utils/utils.py
mohammadnakhaee: Done.

electronicparsers/utils/utils.py
mohammadnakhaee: I changed all of them to TB model

tests/test_tbstudioparser.py
mohammadnakhaee: Done.

tests/test_tbstudioparser.py
mohammadnakhaee: Done.

electronicparsers/tbstudio/parser.py
JosePizarro3: True, sorry, I meant: `self.tbm['grids']['KABC_Coords'].get('value', [])`. Or is the for loop having the same behaviour when looping over None?

mohammadnakhaee: Sorry I knew that it has allways values and it would never be None. I added the get(, []) in the last commit as protection for possible modifications from the tbstudio side

electronicparsers/tbstudio/parser.py
mohammadnakhaee: And another question why the length of species is ["nAtoms"]? Shouldnt it be ["*"]. A fixed length array is wrong I guess.

JosePizarro3: Fixed shapes are useful when you know a priori the length, and there are some checks at the Quantity definition level. But yes, I admit that using `'*'` or `'n_atoms'` is just a matter of taste and not so useful overall.

JosePizarro3:
@mohammadnakhaee let me know when you need another quick round of review. I'll like to merge soon the new refactoring of 'Projection' to 'TB', https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1285. 

Maybe, we should decouple both changes: one merge for the renaming, another merge for adding SlaterKoster and TBStudio.

mohammadnakhaee:
@JosePizarro3 yes. Could you please take a look. I solved the problems and lint errors related to tbstudio but it is still failed. Maybe you know what is going wrong. The failed tests are not related to the tbstudio.

JosePizarro3:
> @JosePizarro3 yes. Could you please take a look. I solved the problems and lint errors related to tbstudio but it is still failed. Maybe you know what is going wrong. The failed tests are not related to the tbstudio.

Maybe rebasing would solve the issue?

Bear in mind that pytest will anyways fail. I'd say that if after rebasing pylint and mypy do not pass, I'll take a look 🙂

mohammadnakhaee:
Yeah let me rebase it. Maybe it helps.

mohammadnakhaee:
I rebased it. Now the problem is only these two

electronicparsers/utils/utils.py:30:0: E0611: No name 'TB' in module 'nomad.datamodel.metainfo.simulation.workflow' (no-name-in-module)
electronicparsers/utils/utils.py:30:0: E0611: No name 'TBMethod' in module 'nomad.datamodel.metainfo.simulation.workflow' (no-name-in-module)

which I guess it makes sense. Doesn't it? They are newly added in the gitlab branch which is not merged yet. Is it correct?

JosePizarro3:
> which I guess it makes sense. Doesn't it? They are newly added in the gitlab branch which is not merged yet. Is it correct?

Yes, indeed 🙂 

Perfect, I will take a look and give you some final review. Though it looked good at first sight. Do you want me also to review now the gitlab side?

mohammadnakhaee:
Yes please see the gitlab again. I will close all the threads since there are extensive changes to the previous version. Would be nice if you comment on this version.

electronicparsers/tbstudio/parser.py
JosePizarro3: Original suggestion:
> Error handling:
>```
>sec_run.program = Program(name='TBStudio', version=self.tbm.get('ReleaseVersion', ''))
>```

The problem is that, if `self.tbm['ReleaseVersion']` is not stored (for whatever reason, like for example a version change), it will return an error. It is safer to do `self.tbm.get('ReleaseVersion')` or `self.tbm.get('ReleaseVersion', '')` if you want to be more verbose with the type of the get.

electronicparsers/tbstudio/parser.py
JosePizarro3: Original suggestion:
> You can:
> ```
> sec_scc = sec_run.m_create(Calculation)
> ```
> 
> And delete the line here where you do `sec_scc = sec_run.calculation[-1]` (see next comment).

Also reopening this suggestion for making `parse_scc` more readable.

electronicparsers/tbstudio/parser.py
JosePizarro3: >You can delete this when doing what I said in the comment before.

Here according to the suggestion of initializing `sec_scc = sec_run.m_create(Calculation)` at the beginning directly.

electronicparsers/tbstudio/parser.py
JosePizarro3: > Error handling (and no need to name 'joules'):
> ```
> fermi_level = self.tbm['variables'].get('ChemP')
> sec_energy = sec_scc.m_create(Energy, Calculation.energy)
> sec_energy.fermi = fermi_level * ureg.eV if fermi_level else None
> ```

And this error handling too

electronicparsers/tbstudio/parser.py
JosePizarro3: > I think you should check if the fermi_level was extracted in order to generate the `BandStructure`. Something like:
> ```
> if not fermi_level:
>    self.logger.warning('Could not extract the Fermi level, so that the BandStructure is not resolved')
>    return
> sec_k_band = BandStructure()
> ...(the rest of your code)
> ```

Here it will be nice, as if for whatever reason `fermi_level` does not exist, the parsing of the bands will be error.

electronicparsers/tbstudio/parser.py
JosePizarro3: >You can delete this too, but otherwise, pi is defined in numpy, as `np.pi`.

This too.

electronicparsers/tbstudio/parser.py
JosePizarro3: Yeah, you can pass astar, bstar, cstar, but I checked that the normalizers take care of this, so you don't need to pass it through the parser 🙂 

Reply: It is strange that I cannot reply to your comments. This is exactly about the 2d material's briluin zone. They are not calculated in the normalizer. So I just wanted to make sure it is done here.

electronicparsers/tbstudio/parser.py
JosePizarro3: > You can deprecate sec_workflow and not use it. Simply do `workflow` and `self.archive.workflow2 = workflow`.

> Please, delete these old lines. Simply add:
> ```
> workflow = SinglePoint()
> self.archive.workflow2 = workflow
> ```
> Everything else is not used anymore. And no need to change the name.

I will keep suggesting this. Do not worry about creating Workflow, as this is an old version that will be deprecated. For new parsers, we don't need to populate it anymore.

electronicparsers/tbstudio/parser.py
JosePizarro3: No need to write the name of the workflow, as this will overwrite `SinglePoint` and we don't want that.

electronicparsers/tbstudio/parser.py
JosePizarro3: > You don't need to open filepath. Simply do `tbm = json.load(filepath)`

This

electronicparsers/tbstudio/parser.py
JosePizarro3: > No need to write this line.

This

electronicparsers/tbstudio/parser.py
JosePizarro3: > No need, do `self.filepath = filepath`. Did you do this because of debugging?

This

electronicparsers/tbstudio/parser.py
JosePizarro3: > Change by `self.archive = archive`.

This

electronicparsers/tbstudio/parser.py
JosePizarro3: > Before defining Run, do a check on loading the json data:
> ```
>         try:
>             tbm = json.load(open(self.filepath))
>         except Exception:
>             self.logger.error('Error opening json output file.')
>             tbm = None
>             return
>         self.tbm = tbm
> ```
> No need to open the filepath.

This

electronicparsers/tbstudio/parser.py
JosePizarro3: > Furthermore, you were written the Program information in `parse_method`. Please, move the program definition here (deleting the corresponding lines in `parse_method`) doing:
> ```
> sec_run.program = Program(name='TBStudio', version=self.tbm.get('ReleaseVersion', ''))
> ```

This, there are two calls for program which are overwritten each other.

electronicparsers/tbstudio/parser.py
JosePizarro3: > No need to define a specific function here. Simply delete that method and do:
> ```
> workflow = SinglePoint()
> self.archive.workflow2 = workflow
> ```

This

electronicparsers/tbstudio/parser.py
JosePizarro3: > No need to call tb_archive, but rather pass `archive` or `self.archive`

This

electronicparsers/utils/utils.py
JosePizarro3: > Is not interesting to also extract the First principles method section so it is referenced to the specific DFT or GW methodology followed?

I think you can check this too. Or leave it for later, as you prefer.

JosePizarro3 (commented):
Hey, just recalling some suggestions and fixes from my previous review.

Furthermore, the data files are not present (they should be in tests/data/tbstudio).

Once these suggestions are implemented, this will be ready 🙂

JosePizarro3:
@mohammadnakhaee all good, come to my office (I have a final doubt) and we do the final merge together 🙂

PR #158, opened 2023-09-11
OpenMX parser: parse more MD details

ondracka:
Just some needed fields so that the new GUI features like the temperature evolution graph are populated correctly.

That said I was not sure how to correctly populate the thermostat details. I just parse the type, but right now it seems it assumes the thermostat is always keeping constant temperature over the whole run, which is not the case at least in the testcases I have. What I have access to from the OpenMX output is either the info about how the target temperature should be at specific timestep (and its linearly interpolated in between), MD trajectory file also actually contains the target interpolated temperature at every timestep. But as I said I was not able to figure out how to put it into the current metainfo.

### DIFF
--- a/electronicparsers/openmx/parser.py
+++ b/electronicparsers/openmx/parser.py
@@ -37,7 +37,7 @@ from nomad.datamodel.metainfo.simulation.method import (
 from nomad.parsing.file_parser import TextParser, Quantity
 from nomad.datamodel.metainfo.simulation.workflow import (
     GeometryOptimization, GeometryOptimizationMethod,
-    MolecularDynamics, MolecularDynamicsMethod)
+    MolecularDynamics, MolecularDynamicsMethod, ThermostatParameters)
 
 from .metainfo.openmx import OpenmxSCC  # pylint: disable=unused-import
 
@@ -96,8 +96,11 @@ mainfile_parser = TextParser(quantities=[
     Quantity('Atoms.UnitVectors.Unit',
              r'(?i)Atoms.UnitVectors.Unit\s+([a-z]{2,3})', repeats=False),
     Quantity('scf.Hubbard.U', r'(?i)scf.Hubbard.U\s+(on|off)', repeats=False),
+    Quantity('MD.maxIter', r'MD\.maxIter\s+(\d+)', repeats=False),
     Quantity('MD.Type', r'(?i)MD\.Type\s+([a-z_\d]{3,6})', repeats=False),
+    Quantity('MD.TimeStep', r'MD\.TimeStep\s+([\d\.e-]+)', repeats=False),
     Quantity('MD.Opt.criterion', r'(?i)MD\.Opt\.criterion\s+([\d\.e-]+)', repeats=False),
+    Quantity('MD.TempControl', r'<MD.TempControl([\s\S]+)MD.TempControl>', repeats=False),
     Quantity('scf.maxIter', r'scf.maxIter\s+(\d+)', repeats=False),
     Quantity('scf.criterion', r'scf.criterion\s+([-\.eE\d]+)', repeats=False),
     Quantity('scf.ElectronicTemperature', r'scf.ElectronicTemperature\s+(\S+)', repeats=False),
@@ -114,6 +117,7 @@ def read_md_file(md_file):
     with open(md_file, 'r') as f:
         cell_vectors_re = re.compile(r'Cell_Vectors=((?:\s+-?\d+\.\d+)+)')
         temperature_re = re.compile(r'Temperature=\s+(\d+\.\d+)')
+        time_re = re.compile(r'time=\s+(\d+\.\d+)')
         for line in f:
             line_list = line.split()
             if len(line_list) == 1:
@@ -133,6 +137,9 @@ def read_md_file(md_file):
                 if temperature is not None:
                     temperature = temperature.group(1)
                     result[-1]['temperature'] = temperature
+                time = time_re.search(line)
+                if time is not None:
+                    result[-1]['time'] = float(time.group(1))
                 step_header = False
             else:
                 result[-1]['positions'][atomindex][0:3] = [
@@ -198,6 +205,40 @@ def parse_structure(system, logger):
         logger.warning('Failed to parse the input structure.')
 
 
+def parse_temperature_profile(data, thermo_type):
+    steps = data[0]
+    thermostats = []
+
+    for i in range(steps):
+        if i == 0 and int(data[1]) == 1:
+            continue
+
+        thermostat = ThermostatParameters()
+        if "VS" in thermo_type:
+            thermostat.thermostat_type = "velocity_rescaling_woodcock"
+            cols = 4
+        if "NH" in thermo_type:
+            thermostat.thermostat_type = "nose_hoover"
+            cols = 2
+
+        if i == 0:
+            thermostat.reference_temperature_start = data[i * cols + 2] * units.K
+            thermostat.temperature_update_frame_start = 1
+        else:
+            thermostat.reference_temperature_start = data[(i - 1) * cols + 2] * units.K
+            thermostat.temperature_update_frame_start = data[(i - 1) * cols + 1]
+        thermostat.reference_temperature_end = data[i * cols + 2] * units.K
+        thermostat.temperature_update_frame_end = data[i * cols + 1]
+
+        if thermostat.reference_temperature_start == thermostat.reference_temperature_end:
+            thermostat.temperature_profile = "constant"
+        else:
+            thermostat.temperature_profile = "linear"
+
+        thermostats.append(thermostat)
+    return thermostats
+
+
 class OpenmxParser:
     def __init__(self):
         pass
@@ -222,9 +263,12 @@ class OpenmxParser:
             workflow = None
             for current_md_type in md_types_list:
                 if current_md_type[0] in md_type:
+                    max_iters = mainfile_parser.get('MD.maxIter')
                     if current_md_type[1] == 'geometry_optimization':
                         workflow = GeometryOptimization(method=GeometryOptimizationMethod())
                         workflow.method.method = current_md_type[2]
+                        if max_iters is not None:
+                            workflow.method.optimization_steps_maximum = max_iters
                         criterion = mainfile_parser.get('MD.Opt.criterion')
                         if criterion is not None:
                             workflow.method.convergence_tolerance_force_maximum = (
@@ -233,8 +277,29 @@ class OpenmxParser:
                             workflow.method.convergence_tolerance_force_maximum = (
                                 1e-4 * units.hartree / units.bohr)
                     else:
-                        workflow = MolecularDynamics(method=MolecularDynamicsMethod())
+                        md_temp_control = mainfile_parser.get('MD.TempControl')
+                        if "VS" in md_type:
+                            thermostats = parse_temperature_profile(md_temp_control, "VS")
+                        elif "NH" in md_type:
+                            thermostats = parse_temperature_profile(md_temp_control, "NH")
+
+                        workflow = MolecularDynamics(
+                            method=MolecularDynamicsMethod(
+                                thermostat_parameters=thermostats))
                         workflow.method.thermodynamic_ensemble = current_md_type[2]
+                        # Everything is saved every time step
+                        workflow.method.coordinate_save_frequency = 1
+                        workflow.method.velocity_save_frequency = 1
+                        workflow.method.force_save_frequency = 1
+                        workflow.method.thermodynamics_save_frequency = 1
+
+                        md_timestep = mainfile_parser.get('MD.TimeStep')
+                        if md_timestep is not None:
+                            workflow.method.integration_timestep = md_timestep * units.fs
+
+                        if max_iters is not None:
+                            workflow.method.n_steps = max_iters
+
             self.archive.workflow2 = workflow
 
     def parse_method(self):
@@ -399,5 +464,10 @@ class OpenmxParser:
                     temperature = mdfile_md_steps[i].get('temperature')
                     if temperature is not None:
                         sec_calc.temperature = temperature * units.kelvin
+                    # Time is also printed for geometry optimizations, but it is meaningless there.
+                    if isinstance(self.archive.workflow2, MolecularDynamics):
+                        time = mdfile_md_steps[i].get('time')
+                        if time is not None:
+                            sec_calc.time = time * units.fs
 
         self.parse_eigenvalues()
--- a/tests/test_openmxparser.py
+++ b/tests/test_openmxparser.py
@@ -124,6 +124,7 @@ def test_AlN(parser):
 
     workflow = archive.workflow2
     assert workflow.method.method == "steepest_descent"
+    assert workflow.method.optimization_steps_maximum == 50
     assert workflow.method.convergence_tolerance_force_maximum.magnitude == approx(
         (0.0003 * units.hartree / units.bohr).to_base_units().magnitude)
 
@@ -182,7 +183,9 @@ def test_C2N2(parser):
     scc = run.calculation
     assert len(scc) == 100
     assert scc[0].temperature.magnitude == approx(300.0)
+    assert scc[0].time == approx(0.0)
     assert scc[99].temperature.magnitude == approx(46.053)
+    assert scc[99].time.magnitude == approx(49.5e-15)
     assert np.shape(scc[0].forces.total.value) == (4, 3)
     assert scc[0].forces.total.value[0][0].magnitude == approx(HaB_to_N(0.10002))
     assert scc[99].forces.total.value[2][2].magnitude == approx(HaB_to_N(-0.00989))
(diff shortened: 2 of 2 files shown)


JosePizarro3:
Hi @ondracka, 

I think it is a good idea to tag @JFRudzinski on this, as he is responsible on extending the MD metainfo and he will happy to help you 🙂

JFRudzinski:
Hi @ondracka 

Sorry for the delayed response.

I am not sure if I understand exactly what you are looking for. It sounds like you want to store the actual system temperatures during a simulation? This is done via archive->run->calculation->temperature, e.g., See [HERE](https://nomad-lab.eu/prod/v1/gui/search/entries/entry/id/hxaepf6x12Xt2IX2jCt4DyfLG0P4/data/run/0/calculation/0/temperature)

We also have some basic support for storing the parameters for a molecular dynamics simulation. This goes under workflow2->method. Here is an example:

```json
{
    "molecular_dynamics": {
        "thermodynamic_ensemble": "NPT",
        "integrator_type": "langevin_leap_frog",
        "integration_timestep": {"value": 2e-15, "unit": "ps"},
        "n_steps": 20000000,
        "coordinate_save_frequency": 10000,
        "velocity_save_frequency": null,
        "force_save_frequency": null,
        "thermodynamics_save_frequency": null,
        "thermostat_parameters": {
            "thermostat_type": "langevin_leap_frog",
            "reference_temperature": {"value": 300.0, "unit": "kelvin"},
            "coupling_constant": {"value": 1.0, "unit": "ps"}},
        "barostat_parameters": {
            "barostat_type": "berendsen",
            "coupling_type": "isotropic",
            "reference_pressure": {"value": [[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]], "unit": "bar"},
            "coupling_constant": {"value": [[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]]},
            "compressibility": {"value": [[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]]}
            }
    }
}
```

I am happy to help further. Just let me know what exactly your calculation/simulation is or what quantities you want to store. It may be the case that we need to extend the metainfo, which is a good thing since we want to start branching this out to cover more scenarios.

We could also Zoom next week sometime if you want for an easier discussion, just let me know here or on the FAIRmat slack if you are there or at joseph.rudzinski@physik.hu-berlin.de.

ondracka:
Thanks for the response, the actual system temperature is already parsed fine, I was looking into the thermostat metainfo, like if I have a NVT run with non-constant temperature, for example something like this is possible with OpenMX: https://www.openmx-square.org/openmx_man3.9/node60.html (but similar temperature profiles can be simulated also with LAMMPS etc.). But from your example it looks like the thermostat metainfo is assuming the temperature is always constant for the whole run (that was also my initial impression).

Other than that I believe everything is clear so the only help I need here is a review and eventual pull ;-)

electronicparsers/openmx/parser.py
```diff
@@ -235,6 +244,30 @@ def parse_workflow(self):
...
+
+                        md_timestep = mainfile_parser.get('MD.TimeStep')
+                        if md_timestep is not None:
+                            workflow.method.integration_timestep = md_timestep * units.fs
+
+                        if max_iters is not None:
+                            workflow.method.n_steps = max_iters
+
+                        # FIXME: The OpenMX docs reference L. V. Woodcock, Chem. Phys. Lett. 10 ,257 (1971).
+                        # for the velocity scaling, while the metainfo "velocity_rescaling" item references
```
JFRudzinski: From a cursory inspection of this reference, it appears to be a more direct approach, where the velocities are rescaled deterministically. Whereas the Bussi algorithm is stochastic. I am not an expert in the history of these algorithms, but I would suggest that we add another thermostat_type called "velocity_rescaling_woodcock" and then reference this paper (I will open an issue and branch for this in the NOMAD-lab repo and provide you the link). 

If you have any other further information/thoughts about this please share. I want to be precise about which algorithm is being used, but I do not want to just list a million different methods with accompanying papers, if the methods are essentially the same.

electronicparsers/openmx/parser.py
```diff
@@ -399,5 +432,10 @@ def parse(self, mainfile: str, archive: EntryArchive, logger):
                     temperature = mdfile_md_steps[i].get('temperature')
                     if temperature is not None:
                         sec_calc.temperature = temperature * units.kelvin
+                    # Time is also printed for geometry optimizations, but it is meaningless there.
```
JFRudzinski, line 467: I think we should probably not store the time quantity for geometry optimizations. I know this is also currently done for Gromacs and I have a note to change it. But I need to make sure it is consistent across the board.

@ladinesa Can we safely simply not store the time quantity for geometry optimizations?

electronicparsers/openmx/parser.py
JFRudzinski: Am updating the metainfo now, see https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1466 

once I merge you can use "velocity_rescaling_woodcock" for the thermostat_type

JFRudzinski (approved):
@ondracka Overall it looks good to me.

As you will see in my comments, I am addressing your issue with the thermostat_type.

I also went ahead and added repeats to the thermostat and barostat sections, along with quantities `frame_start` and `frame_end`. This way you can add something like a simulated annealing protocol, where you change the target temperature throughout the simulation. See also here: https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1466

Can you let me know if this would suit your needs? Once you confirm, I will merge and then you will need to simply make thermostat_parameters a list. I would suggest then using:

```
            workflow = MolecularDynamics(
                method=MolecularDynamicsMethod(
                    thermostat_parameters=[ThermostatParameters()],
                    barostat_parameters=[BarostatParameters()]
                ), results=MolecularDynamicsResults()
            )
```
to define the workflow section instead of using m_create()

ondracka:
> @ondracka Overall it looks good to me.
> 
> As you will see in my comments, I am addressing your issue with the thermostat_type.
> 
> I also went ahead and added repeats to the thermostat and barostat sections, along with quantities `frame_start` and `frame_end`. This way you can add something like a simulated annealing protocol, where you change the target temperature throughout the simulation. See also here: https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1466
> 
> Can you let me know if this would suit your needs? Once you confirm, I will merge and then you will need to simply make thermostat_parameters a list. I would suggest then using:
> 
> ```
>             workflow = MolecularDynamics(
>                 method=MolecularDynamicsMethod(
>                     thermostat_parameters=[ThermostatParameters()],
>                     barostat_parameters=[BarostatParameters()]
>                 ), results=MolecularDynamicsResults()
>             )
> ```
> 
> to define the workflow section instead of using m_create()


Yeah, I think this should alow to store any crazy temperature profiles whatsoever.
However I'm not sure I follow regarding not using the `m_create()`. For example if I have a simulated heating from 300K to 400K with a constant heating rate over a 100 timesteps (so 1K/timestep), it means I'll have thermostat_parameters section repeat 100 times (each one with the actual interpolated target temperature for the specific MD step and where `frame_start` and `frame_end` will be the same single frame where this specific target temperature is relevant.) But I still have only one workflow, so I still need to use m_create() for the thermostat_parameters sections don't I?

electronicparsers/openmx/parser.py
ondracka: Just to make sure, the current PR is fine as is here right (with not saving the time)? I think it is mostly printed only because OpenMX uses same output template for MD and geometry optimization.

That said other DFT codes can have some hybrid stuff (like damped MD geometry optimization) which kind of falls in between of the two...

ondracka:
Thanks for the comments, I'll send an updated PR with the comments resolved when https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1466 is merged

JFRudzinski:
> > @ondracka Overall it looks good to me.
> > As you will see in my comments, I am addressing your issue with the thermostat_type.
> > I also went ahead and added repeats to the thermostat and barostat sections, along with quantities `frame_start` and `frame_end`. This way you can add something like a simulated annealing protocol, where you change the target temperature throughout the simulation. See also here: https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1466
> > Can you let me know if this would suit your needs? Once you confirm, I will merge and then you will need to simply make thermostat_parameters a list. I would suggest then using:
> > ```
> >             workflow = MolecularDynamics(
> >                 method=MolecularDynamicsMethod(
> >                     thermostat_parameters=[ThermostatParameters()],
> >                     barostat_parameters=[BarostatParameters()]
> >                 ), results=MolecularDynamicsResults()
> >             )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to define the workflow section instead of using m_create()
> 
> Yeah, I think this should alow to store any crazy temperature profiles whatsoever. However I'm not sure I follow regarding not using the `m_create()`. For example if I have a simulated heating from 300K to 400K with a constant heating rate over a 100 timesteps (so 1K/timestep), it means I'll have thermostat_parameters section repeat 100 times (each one with the actual interpolated target temperature for the specific MD step and where `frame_start` and `frame_end` will be the same single frame where this specific target temperature is relevant.) But I still have only one workflow, so I still need to use m_create() for the thermostat_parameters sections don't I?

In that case you could do something like:
```
            n_targets = 100
            thermostat_section = [Thermostat_Parameters() for _ in range(n_targets)]
            workflow = MolecularDynamics(
                method=MolecularDynamicsMethod(
                    thermostat_parameters=thermostat_section,
                    barostat_parameters=[BarostatParameters()]
                ), results=MolecularDynamicsResults()
            )

            # populate
            for i_thermo in range(n_targets):
                workflow.method.thermostat_parameters[i_thermo].reference_temperature = xxxx
```

However, now that you gave this example, I think I still want to improve the metainfo. For some reason I wasn't really thinking about a continuous protocol. I am thinking that we should also give the option to simply give the starting/stopping frames and then also provide a type of interpolation (e.g., linear), and the interpolation rate. Then you could store everything in a single section if you want (we will still allow multiple sections for more complicated protocols). What do you think? Are there other quantities that I should define?

electronicparsers/openmx/parser.py
ladinesa: > I think we should probably not store the time quantity for geometry optimizations. I know this is also currently done for Gromacs and I have a note to change it. But I need to make sure it is consistent across the board.
> 
> @ladinesa Can we safely simply not store the time quantity for geometry optimizations?

I am actually interested in wall clock times: time_physical and time_calculation not time (simulation time).

ondracka: What is available in OpenMX output is starting wallclock time and than elapsed time at the end. There is also elapsed time per MD/geometry optimization step printed to stdout (which might or might not be saved somewhere in addition to the mainfile), but how precise this could be to reconstruct the wallclock time at every step is unknown (i.e., if summing the partial elapsed times will sum to the total elapsed time of the run).

I can add support to the parser but that seems unrelated to this PR.

ladinesa: > What is available in OpenMX output is starting wallclock time and than elapsed time at the end. There is also elapsed time per MD/geometry optimization step printed to stdout (which might or might not be saved somewhere in addition to the mainfile), but how precise this could be to reconstruct the wallclock time at every step is unknown (i.e., if summing the partial elapsed times will sum to the total elapsed time of the run).
> 
> I can add support to the parser but that seems unrelated to this PR.

I can also have a look later as I am doing this for all the parsers at the moment. I need this info to construct the workflow tasks.

ondracka:
Most codes I know can do only a linear heating/cooling between starting and ending temperature (this is also what LAMMPS has https://docs.lammps.org/fix_nh.html or equivalently starting and ending pressure for the barostat or IIRC also VASP has similar syntax with TEBEG and TEEND).

AFAIK Qunatum ESPRESSO sets starting temperature and instead of the final temperature it sets how often and how is the target temperature changed (either by adding some constant value which would be equivalent to the linear interpolation, except one would have to calculate the final temperature by hand based on the number of steps and the increment however it also allows to rescale/multiply the current target temperature by fixed amount which would be equivalent to exponential heating cooling profile https://www.quantum-espresso.org/Doc/INPUT_PW.html#delta_t and https://www.quantum-espresso.org/Doc/INPUT_PW.html#nraise)

So I think your suggestion would work for me (and is much more breve than the initial suggestion) however what you proposed initially is probably most generic.

JFRudzinski:
> Most codes I know can do only a linear heating/cooling between starting and ending temperature (this is also what LAMMPS has https://docs.lammps.org/fix_nh.html or equivalently starting and ending pressure for the barostat or IIRC also VASP has similar syntax with TEBEG and TEEND).
> 
> AFAIK Qunatum ESPRESSO sets starting temperature and instead of the final temperature it sets how often and how is the target temperature changed (either by adding some constant value which would be equivalent to the linear interpolation, except one would have to calculate the final temperature by hand based on the number of steps and the increment however it also allows to rescale/multiply the current target temperature by fixed amount which would be equivalent to exponential heating cooling profile https://www.quantum-espresso.org/Doc/INPUT_PW.html#delta_t and https://www.quantum-espresso.org/Doc/INPUT_PW.html#nraise)
> 
> So I think your suggestion would work for me (and is much more breve than the initial suggestion) however what you proposed initially is probably most generic.

ok, great. Please check what I added here: https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/blob/1684-new-thermostat-metadata-needed-for-openmx-parsing-of-md-data/nomad/datamodel/metainfo/simulation/workflow.py

If you have any suggested adjustments to naming or descriptions, please let me know.

Also, I forgot to mention before, you should add tests for all your changes, I did not see any during the review. Then, you can also add tests for this new metainfo, which is needed.

ondracka:
It looks fine (that said my opinion has no weight whatsoever), however right now one would need to do some recalculations in the parsers to fill this (from starting and ending temperature to the `annealing_delta` in OpenMX and LAMMPS, etc..). Aren't the parsers supposed to just get what is in the output files and the rest should be done in normalizing?

So it might be cleaner to add both `reference_temperature_start` (this would be also the reference temperature for the constant profile) and `reference_temperature_end`. Annealing_type I would rename to `temperature_profile` and I would add also "constant" option for the simple case. Also frame_end and frame_start should be valid also for the constant temperature (in lot of cases users do thermalizing at constant temperature first and than some heating/cooling).
Also if the annealing_type is renamed than the others could also be renamed to something like
`temperature_update_frequency`, `temperature_update_delta/increment`, `temperature_update_scaling_factor`, etc...

So this would allow for every parser to just parse its native inputs, like
- constant T only
- Tstart frame_start Tend  frame_end combination
-  Tstart + temperature_update_increment + temperature_update_frequency, etc...

And than normalizer would later recalculate it to fill the other fields like reference_temperature_end or the increment/factor (and potentially to calculate specific target temperature/pressure at specific timesteps, so it could be potentially plotted together with the actual values in the "Thermodynamic properties" box on the entry page similarly to the OpenMX manual plots I referenced earlier).

JFRudzinski:
> And than normalizer would later recalculate it to fill the other fields like reference_temperature_end or the increment/factor (and potentially to calculate specific target temperature/pressure at specific timesteps, so it could be potentially plotted together with the actual values in the "Thermodynamic properties" box on the entry page similarly to the OpenMX manual plots I refere

Great, thank you for the suggestions! I like your naming, I wasn't happy with annealing but wasn't sure what to call it.

The only part I disagree with is the last part about the Thermodynamic properties. What we are talking about are parameters that are part of the MD workflow. The temperature quantity in Thermodynamic properties is the actual simulation temperature (i.e., calculated via the velocities), not the target temperature. That being said, in the future we could have plots or widgets for plotting quantities for common workflows, and in this case we could have something like what you are talking about.

Also, indeed as you said, we need to do a lot of reprocessing and, in the case of Lammps, a lot of improvements in converting the native quantities to the normalized metainfo. We are working on this little by little depending on demand essentially.

Anyway, this is great, I will make these changes and let you know when everything is merged.

JFRudzinski:
@ondracka Sorry for the delay! It turned out we had to fix a couple things for the new metainfo to work properly. I just merged the NOMAD branch mentioned above into develop. Let me know if you need anything else.

ondracka:
> @ondracka Sorry for the delay! It turned out we had to fix a couple things for the new metainfo to work properly. I just merged the NOMAD branch mentioned above into develop. Let me know if you need anything else.

Thanks, I'll try to rebase tomorrow.

electronicparsers/openmx/parser.py
ondracka: OK,  updated for the `velocity_rescaling_woodcock`

ondracka:
This is now rebased on top of the thermostat metainfo changes.

ondracka:
Hm, I see some test failures, it looks like the some other parsers (at least FHIaims) need to be update to the thermostat_parameters metainfo changes (the OpenMX tests all pass properly locally).

JFRudzinski:
> Hm, I see some test failures, it looks like the some other parsers (at least FHIaims) need to be update to the thermostat_parameters metainfo changes (the OpenMX tests all pass properly locally).

Ah yes, this should be very easy, we just need to change the thermostat section to a list. I opened an issue here: https://github.com/nomad-coe/electronic-parsers/issues/167, However, I am helping to run a workshop this week and may not have time to fix it. I would say you can move forward with this PR though, and then I will fix these tests when I return.

ondracka:
@ladinesa @JFRudzinski I need someone to push the merge button here, I  believe all the issues were resolved, but I don't have the rights myself.

JFRudzinski:
@ondracka sorry about the delay! I hope everything is working for you now.

ondracka:
> @ondracka sorry about the delay! I hope everything is working for you now.

Np, all is fine, thanks for the review.

PR #160, opened 2023-09-12
Fix error with `silicon_versions` fixture and `parse_native_tiers`

### LINKED ISSUE #145: FHI-aims testing error
@ladinesa @ndaelman-hu I am trying to run locally the tests (using `python -m pytest -sv tests`) and it seems that FHI-aims has an error. Here is the message I got:

```
============================================================================================================================== ERRORS ==============================================================================================================================
___________________________________________________________________________________________ ERROR collecting dependencies/parsers/electronic/tests/test_fhiaimsparser.py ___________________________________________________________________________________________
Fixture "silicon_versions" called directly. Fixtures are not meant to be called directly,
but are created automatically when test functions request them as parameters.
See https://docs.pytest.org/en/latest/fixture.html for more information about fixtures, and
https://docs.pytest.org/en/latest/deprecations.html#calling-fixtures-directly about how to update your code.
===================================================================================================================== short test summary info ======================================================================================================================
ERROR tests/test_fhiaimsparser.py
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

Can any of you take a look on this 🙂? Pylens is getting crazy so I am not sure who did these changes.

ndaelman-hu:
Resolve breaking tests `test_band_silicon` and `test_native_tiers`

Closes #145


tests/test_fhiaimsparser.py
```diff
@@ -53,12 +49,10 @@ def silicon(parser, silicon_versions):
 
 
 @pytest.fixture(scope='module')
-def silicon_normalization_factors(silicon_versions):
-    normalization_factors = [.5, 1]
-    return dict(zip(silicon_versions, normalization_factors))
+def silicon_normalization_factors():
```
ladinesa: no need to make this fixture as used only in one test. simply define this locally in the test.

tests/test_fhiaimsparser.py
```diff
@@ -37,12 +38,7 @@ def parser():
 
 
 @pytest.fixture(scope='module')
-def silicon_versions():
-    return ('v071914_7', 'v171221_1')
-
-
-@pytest.fixture(scope='module')
-def silicon(parser, silicon_versions):
+def silicon(parser):
```
ladinesa, line 42: no need to make it fixture.

tests/test_fhiaimsparser.py
```diff
@@ -53,12 +49,10 @@ def silicon(parser, silicon_versions):
...
 @pytest.fixture(scope='module')
-def silicon_normalization_factors(silicon_versions):
-    normalization_factors = [.5, 1]
-    return dict(zip(silicon_versions, normalization_factors))
+def silicon_normalization_factors():
+    return dict(zip(silicon_versions, [.5, 1]))
 
 
-@pytest.fixture(scope='module')
 def parse_native_tiers(tier):
```
ladinesa: no need to make separate function put it inside test.

ladinesa:
Please also include a smaller test data in this pr.

tests/test_fhiaimsparser.py
ndaelman-hu: Here neither? `silicon` is used 2x, and as a `module` fixture it only parses both versions once.
Moreover, `silicon` was a fixture even before I added both versions.

ladinesa: oh ok i miss the other one.

ndaelman-hu:
> Please also include a smaller test data in this pr.

I'm generating examples myself now. These should be single point and be much smaller.

ndaelman-hu:
Done, I removed the Hessian files and made `aims.out` 6x smaller. `intermediate` still takes a bit, but overall the impact is noticeable. After updating the native tier in the fhiaims, I could also just go with `control.in`, which would speed it up way better.

ladinesa:
> Done, I removed the Hessian files and made `aims.out` 6x smaller. `intermediate` still takes a bit, but overall the impact is noticeable. After updating the native tier in the fhiaims, I could also just go with `control.in`, which would speed it up way better.

If this is the smallest you can generate then ok. We want to keep each test file under 1mb as much as possible.

ladinesa:
Please merge.

PR #163, opened 2023-09-13
Add calculation time

ladinesa:


ladinesa:
@JosePizarro3  I am simply adding calculation times. I need this data in constructing workflow tasks. Would be great if you have a look at it quickly. thanks.

electronicparsers/cpmd/parser.py
```diff
@@ -419,6 +423,7 @@ def write_energies(source, target):
                 if len(trajectory[n_frame][0]) > 1:
                     sec_system.atoms.velocities = trajectory[n_frame, :, 1, :] * ureg.bohr / (ureg.dirac_constant / ureg.hartree)
 
+                time_initial = sec_run.calculation[-1].time_physical if n_frame else 0 * ureg.s
```
JosePizarro3, line 426: What happens if you don't multiply by ureg.s? Does the metainfo store `0s` or simply ignoring the quantity and consider it not existing?

electronicparsers/cpmd/parser.py
```diff
@@ -431,6 +436,7 @@ def write_energies(source, target):
                             start = energies_n
                             write_energies(energies[energies_n], sec_calc)
                             break
+                sec_calc.time_physical = sec_calc.time_calculation + time_initial
```
JosePizarro3, line 439: Maybe you can move time_initial on top of this? Just to have them nearby 🙂

electronicparsers/crystal/parser.py
```diff
@@ -925,6 +926,8 @@ def add_functionals(functionals):
                 i_system = system
                 i_energy = steps[0]["energy"]
                 scc.energy.total = EnergyEntry(value=i_energy)
+                scc.time_physical = steps[0]["time_physical"]
+                scc.time_calculation = steps[0]["time_physical"]
```
JosePizarro3, line 931: Wait, I am confused now. I thought `time_physical` refers to the time at which the scf/calculation was done, while `time_calculation` is the relative time to perform a scf/calculation. These 2 lines confuse me.

electronicparsers/orca/metainfo/orca.py
```diff
@@ -1085,6 +1085,12 @@ class Calculation(simulation.calculation.Calculation):
         description='''
         ''')
 
+    x_orca_scf_gradient = Quantity(
```
JosePizarro3: What is this quantity? I would delete it and not store it under calculation, as you are saving it under `time_calculation` and `time_physical` already.

tests/test_castepparser.py
```diff
@@ -58,7 +58,8 @@ def test_single_point(parser):
     assert len(sec_scfs) == 12
     assert sec_scfs[3].energy.total.value.magnitude == approx(-2.21530505e-16)
     assert sec_scfs[8].energy.fermi.magnitude == approx(8.53007633e-19)
-    assert sec_scfs[6].time_calculation.magnitude == 12.70
+    # uncomment one time_physical def is updated
+    # assert sec_scfs[6].time_physical.magnitude == 12.70
```
JosePizarro3, line 62: These and other testings should be uncomment, right?

JosePizarro3 (commented):
Hi,

Very nice. Just a few minor suggestions, and I have a couple of minor doubts:

1. I think you could add testing to all parsers you touch, thought I perfectly understand this is a lengthy and boring process. But it will help us to maintain `time_physical` and `time_calculation`.
2. Maybe you can improve the descriptions in the `time_physical` and `time_calculation` definitions in datamodel? It took me a bit to understand them, so I suggest to rewrite them slightly + add some examples, so the people who wants to use them understand what are the differences.

Thanks!

electronicparsers/cpmd/parser.py
ladinesa: It would still be 0s, the reason I am multiplying it here is that additing pint quantities only works for same units.

electronicparsers/cpmd/parser.py
ladinesa: time initial should be assigned before creating a new calculation section

electronicparsers/crystal/parser.py
ladinesa: this is the first frame so time_physical = time_calculation.

electronicparsers/orca/metainfo/orca.py
ladinesa: this is the time only for scf, not necessarily the calc time but for relaxation step, it is unfortunately not printed so this is an approximation.. I can also not simply assign scf gradient time  for the relaxation step.

tests/test_castepparser.py
ladinesa: Yes I will just update time_physical, it is integer right now so it fails test.

electronicparsers/cpmd/parser.py
JosePizarro3: I see, then nevermind.

electronicparsers/crystal/parser.py
JosePizarro3: Ok, understood, thanks! 🙂

electronicparsers/orca/metainfo/orca.py
JosePizarro3: Yeah, my point is that it is not being assigned anywhere in the parsing, right?

JosePizarro3:
Ok, everything's great. 

Maybe you can discuss with Joseph about times under calculation. I think right now we have 4 quantities that have slightly different meanings, but still the descriptions are not proper. Furthermore, what about `run.time_run`? These quantities should be related with those defined there.

electronicparsers/orca/metainfo/orca.py
ladinesa: removed metadata

PR #164, opened 2023-09-15
Fix solid dmft

JosePizarro3:
Fixing the solid_dmft parser, and taking away things that made the archive very heavy.

Now parsing also supports real frequencies and DOS parsing.


JosePizarro3:
Hi @ladinesa these are some changes on the solid_dmft parser. An user pointed out some bugs, and I took this chance to improve on the parser, moving heavy stuff to JSON types and HDF5Reference. 

I tested it in several examples and it seems to be working nicely. But feel free to suggest changes please 🙂

electronicparsers/soliddmft/parser.py
```diff
@@ -127,238 +117,396 @@ def parse_system(self):
         else:
             pass
 
-    def parse_input_model(self, data):
+    def parse_input_model(self, data: h5py.File):
```
ladinesa, line 131: you are actually passing the whole h5 data here, but a bit down, you also use self.dft_input, maybe pass only the groups that are relevant?

electronicparsers/soliddmft/parser.py
```diff
@@ -127,238 +117,396 @@ def parse_system(self):
...
+                continue
+            params = self.parse_groups_datasets(group)
+            sec_method.m_set(sec_method.m_get_quantity_definition(f'x_soliddmft_{group_name}'), params)
 
         # HoppingMatrix || ProjectionMatrix
-        sec_hamiltonian.projection_matrix = self.dft_input.get('proj_mat')[()][:, 0, :, :, :, 0] \
-            + self.dft_input.get('proj_mat')[()][:, 0, :, :, :, 1] * 1j
+        proj_mat = self.dft_input.get('proj_mat')
+        if proj_mat:
+            proj_mat = proj_mat[()]
```
ladinesa: for all of these, can you use parse_groups_datasets (maybe parse_h5) so you do not need [()] and the additional if, if you return None for non-existent dataset.

electronicparsers/soliddmft/parser.py
```diff
@@ -127,238 +117,396 @@ def parse_system(self):
...
-
-            sec_hubbard_kanamori_model.orbital = self.angular_momentum[
-                self.dft_input['corr_shells'][str(i)].get('l', 2)[()]]
-            sec_hubbard_kanamori_model.u = self.dmft_input['general_params']['U'][str(i)][()] * ureg.eV
-            sec_hubbard_kanamori_model.jh = self.dmft_input['general_params']['J'][str(i)][()] * ureg.eV
-            # solid_dmft keeps spin-rotational invariance
-            sec_hubbard_kanamori_model.up = sec_hubbard_kanamori_model.u - 2.0 * sec_hubbard_kanamori_model.jh
+        n_impurities = self.dft_input.get('n_inequiv_shells')[()] if self.dft_input.get('n_inequiv_shells') else 1
+        general_params = self.dmft_input.get('general_params')
+        for i in range(n_impurities):
```
ladinesa: i would use n here, and maybe do only one str conversion and not use f formatting?.

electronicparsers/soliddmft/parser.py
```diff
@@ -127,238 +117,396 @@ def parse_system(self):
...
+            sec_atom_parameters = sec_method.m_create(AtomParameters)
+            atom_impurity = self.dft_input.get('corr_shells')
+            if atom_impurity:
+                atom_index = atom_impurity.get(f'{i}').get('atom')
```
ladinesa: error handling necessary here?

electronicparsers/soliddmft/parser.py
```diff
@@ -127,238 +117,396 @@ def parse_system(self):
...
+                n_orbitals = atom_impurity.get(f'{i}').get('dim')
+                l_number = atom_impurity.get(f'{i}').get('l')
+                if n_orbitals and l_number:
+                    sec_atom_parameters.n_orbitals = n_orbitals[()]
+                    angular_momentum = self.angular_momentum[l_number[()]]
+                    orbital_labels = [f'{angular_momentum}{ml}' for ml in range(n_orbitals[()])]
+                    sec_atom_parameters.orbitals = orbital_labels
+
+            sec_hubbard_kanamori_model = sec_atom_parameters.m_create(HubbardKanamoriModel)
+            if general_params.get('U')[f'{i}'] and general_params.get('J')[f'{i}']:
```
ladinesa: error handling?

ladinesa: maybe split the if for U and J, nvm you need both for up

electronicparsers/soliddmft/parser.py
```diff
@@ -127,238 +117,396 @@ def parse_system(self):
...
-        sec_dmft.inverse_temperature = beta / ureg.eV
-        if sec_method.x_soliddmft_general.x_soliddmft_magnetic:
-            if sec_method.x_soliddmft_general.x_soliddmft_magmom is None:
-                self.logger.warning('The magnetic flag is set to true, but the initial magnetic moment is not resolved. '
-                                    'Is this really a magnetic calculation without an initial magmom seed?')
-            else:
-                if all(signs == 1.0 for signs in np.sign(sec_method.x_soliddmft_general.x_soliddmft_magmom)):
+        n_impurities = self.dft_input.get('n_inequiv_shells')
+        sec_dmft.n_impurities = n_impurities[()] if n_impurities else 1
+        try:
```
ladinesa: what could go wrong here? not possible to remove try?

electronicparsers/soliddmft/parser.py
```diff
@@ -127,238 +117,396 @@ def parse_system(self):
...
+        beta = dmft_general_params.get('beta')
+        if beta:
+            beta = beta[()]
+            sec_dmft.inverse_temperature = beta / ureg.eV
+        magnetic_flag = dmft_general_params.get('magnetic')[()] if self.dmft_input.get('general_params', {}).get('magnetic') else False
+        if magnetic_flag:
+            magmom = dmft_general_params.get('magmom')
+            if len(magmom) > 0:
+                magmom = [magmom.get(keys)[()] for keys in magmom.keys()]
+                if all(signs == 1.0 for signs in np.sign(magmom)):
```
ladinesa: you can do this in one line.

electronicparsers/soliddmft/parser.py
```diff
@@ -127,238 +117,396 @@ def parse_system(self):
...
+                i_scf (int): Scf iteration index.
+                n_impurities (int): Number of impurities.
+                convergence_obs (Dict[str, Any]): Convergence of observables.
+                observables (Dict[str, Any]): Observables.
+            """
+            sec_scf = sec_scc.m_create(ScfIteration)
+            # Convergence of the observables per scf iteration step
+            conv_obs = {}
+            for key, value in convergence_obs.items():
+                if len(value) > 0 and key != 'iteration':
```
ladinesa: i prefer if
```python
if len(value) == 0 and key == 'iteration':
    continue

File: electronicparsers/soliddmft/parser.py [hunk 9]
```diff
@@ -127,238 +117,396 @@ def parse_system(self):
...
+        def parse_gfs(sec_scc: Calculation, sec_gfs: GreensFunctions, n_impurities: int):
+            """Parses Green's functions quantities.
+
+            Args:
+                sec_scc (Calculation): Calculation section.
+                sec_gfs (GreensFunctions): Green's function section under Calculation.
+                n_impurities (int): Number of impurities.
+            """
+
+            def parse_dos(sec_scc: Calculation, sec_gfs: GreensFunctions, chemical_potential: np.float64):
```
ladinesa: I am also guilty of this but nesting functions can make it hard to read. do it when really necessary.

electronicparsers/soliddmft/parser.py
```diff
@@ -127,238 +117,396 @@ def parse_system(self):
...
+                """Parses DOS if the impurity solver returns Green's functions in the real
+                frequency space (MPS).
+
+                Args:
+                    sec_scc (Calculation): Calculation section.
+                    sec_gfs (GreensFunctions): Green's functions section
+                    chemical_potential (np.float64): Magnitude of the chemical potential in eV.
+                """
+                magnetic_state = sec_scc.method_ref.dmft.magnetic_state
```
ladinesa: condense to one liner

electronicparsers/soliddmft/parser.py
```diff
@@ -127,238 +117,396 @@ def parse_system(self):
...
+
+            if sec_scc.method_ref.dmft.n_correlated_orbitals is not None:
+                for gf_quantity in self.iteration_gfs:
+                    quantity_name = axes.get(gf_quantity)
+                    if not quantity_name:
+                        continue
+                    value_per_atom = []
+                    for i_imp in range(n_impurities):
+                        n_orb = sec_scc.method_ref.dmft.n_correlated_orbitals[i_imp]
+                        if dmft_last_iter.get(f'{gf_quantity}_{i_imp}'):
```
ladinesa: use continue here to avoid extra indent

electronicparsers/soliddmft/parser.py
```diff
@@ -127,238 +117,396 @@ def parse_system(self):
...
+        if len(scf_keys) != n_iter_scf:
+            self.logger.warning('Number of it_n steps for DMFT results is different from the iteration '
+                                'count stored in the h5 file.')
 
-        # last iteration quantities
-        parse_iteration_quantities(sec_scf_iteration, 'last_iter')
+        n_impurities = self.dft_input.get('n_inequiv_shells')[()]
+        convergence_obs = self.dmft_results.get('convergence_obs', {})
+        observables = self.dmft_results.get('observables', {})
+        if self.archive.m_context:
```
ladinesa: sorry I do not follow, why need to access file here?

electronicparsers/soliddmft/parser.py
```diff
@@ -127,238 +117,396 @@ def parse_system(self):
...
-                naxis = 2 * n_iw
-            else:
-                naxis = n_tau
-            if np.all(norb != norb[0]):
-                self.logger.warning('Greens function matrices are set up using the number of orbitals from the impurity 0. '
-                                    'We found different number of orbitals per impurity. Is this physically correct?')
-            setattr(sec_gf, self._gf_map[keys], np.reshape(funct, (nat, 2, norb[0], naxis)))
-        sec_gf.orbital_occupations = np.reshape(sec_scc.scf_iteration[-1].x_soliddmft_observables.x_soliddmft_orb_occ, (nat, 2, norb[0]))
-        sec_gf.quasiparticle_weights = np.reshape(sec_scc.scf_iteration[-1].x_soliddmft_observables.x_soliddmft_orb_Z, (nat, 2, norb[0]))
+        sec_gfs = sec_scc.m_create(GreensFunctions)
```
ladinesa: maybe put section creation inside function and return it, if you use it later just to be consistent .

electronicparsers/utils/utils.py
```diff
@@ -85,6 +86,24 @@ def get_files(pattern: str, filepath: str, stripname: str = '', deep: bool = Tru
     return filenames
 
 
+def numpy_type_to_json_serializable(quantity: Union[np.bool_, np.int32, np.int64, np.float64]):
```
ladinesa, line 89: maybe try to make a custom NumpyJsonEncoder class instead of function.

electronicparsers/soliddmft/parser.py
JosePizarro3: In line 134 I am looping over some groups in data. I think in this function there is no other way than passing 3 ('dft_input', 'dft_misc_input', 'dft_symmcorr_input') out of 5 groups (+ 'DMFT_input', 'DMFT_results') from the h5 file.

electronicparsers/soliddmft/parser.py
JosePizarro3: I'll prefer to keep n_impurities, but yes, I will do the str conversion instead of f.

electronicparsers/soliddmft/parser.py
JosePizarro3: Sorry, I understood that `to_hdf5` works as HDF5Reference only if the archive has m_context, but maybe not? Just copied the lines from the VASP parser.

About opening the file, I forgot to add it in `parse_scf`, but I want to point to that specific file `f`. I will fix this.

electronicparsers/utils/utils.py
JosePizarro3: Even for this simple return? I think a function is easier to maintain, and I doubt I am going to extend this. My argument is that this should be directly in `metainfo.py`.

electronicparsers/soliddmft/parser.py
ladinesa: ok, you can also pass the whole data then.

electronicparsers/soliddmft/parser.py
ladinesa: i actually meant n instead of i but no big deal.

electronicparsers/utils/utils.py
ladinesa: ok i agree.

electronicparsers/soliddmft/parser.py
JosePizarro3: Ah sorry, ok, I agree with you 👍🏻

electronicparsers/soliddmft/parser.py
JosePizarro3: ~~Yeah, it might be cases where corr_shells are not reported.~~ Ah sorry, you meant for the string `f'{i}'`. No need I'd say, if it does not exist then there is some problem with the output.

electronicparsers/soliddmft/parser.py
JosePizarro3: Ok, I defined `_extract_section` to handle these situations. Please, check it out. I also add a default value for some specific cases.

electronicparsers/soliddmft/parser.py
JosePizarro3: Ok, done.

electronicparsers/soliddmft/parser.py
JosePizarro3: (see above)

electronicparsers/soliddmft/parser.py
JosePizarro3: It is indeed. Now with the new `_extract_dataset` it should work better and try is not necessary.

electronicparsers/soliddmft/parser.py
JosePizarro3: Ok, done.

electronicparsers/soliddmft/parser.py
JosePizarro3: Yeah, me too 😄  Done, check it out.

electronicparsers/soliddmft/parser.py
JosePizarro3: Ok, I deleted this nested function.

electronicparsers/soliddmft/parser.py
JosePizarro3: Done.

electronicparsers/soliddmft/parser.py
JosePizarro3: Done.

electronicparsers/soliddmft/parser.py
JosePizarro3: Done.

JosePizarro3:
Thanks a lot for the review. I had to also fix some GW parsing, as now `FrequencyMesh` is a repeated quantity.

Let me know what you think.

electronicparsers/soliddmft/parser.py
```diff
@@ -68,297 +58,457 @@ def __init__(self):
...
+        Args:
+            dataset (Optional[h5py.Dataset]): H5py dataset to be extracted.
+            default (Union[bool, int, float, np.ndarray, None], optional): Default value. Defaults to None.
+
+        Returns:
+            Union[np.bool_, np.int32, np.int64, np.float64, np.ndarray]: Returns the value associated with the dataset.
+        """
+        return dataset[()] if dataset else default
+
+    def parse_groups_datasets(self, group: h5py.Group):
```
ladinesa: I would actually combine these two functions that works for any h5 element.

electronicparsers/soliddmft/parser.py
ladinesa: i suggest you pass the filename rather thatn the context and access only the file at the part where you need to.

electronicparsers/soliddmft/parser.py
JosePizarro3: Not sure if I understand. 

`parse_group_datasets` reads the group and saves it into a dictionary of {key: val} for a type=JSON quantity.

`_extract_dataset` simply returns the dataset or default.

Maybe you can explain me more your idea?

ladinesa: you can simply return the value of the dataset at the start of the function.

JosePizarro3: Ok, check it out now, I defined `extract_groups_datasets`.

electronicparsers/soliddmft/parser.py
JosePizarro3: Ok, I am passing `filename` and `farg` to `parse_scf` now.

ladinesa: maybe not pass farg as it is a constant.

JosePizarro3: True. Ok, done.

JosePizarro3:
If approved, I can merge this and contact the developer back. Thanks for the review, @ladinesa 🙂

PR #166, opened 2023-09-19
Refactor extract_section

JosePizarro3:
A minor change to import `extract_section` directly from the central nomad.utils.


JosePizarro3:
Closes #167

PR #176, opened 2023-11-01
Modify text parser

ladinesa:
Adapted several parsers to changes in DataTextParser


JosePizarro3 (changes requested):
Looks good to me, but you didn't adapt the `DataANDTextParser` from the utils.py. This is only used for the edmft parser, can you test it works as intended with the test data example?

ladinesa:
Ah yes, 

> Looks good to me, but you didn't adapt the `DataANDTextParser` from the utils.py. This is only used for the edmft parser, can you test it works as intended with the test data example?

Ah sorry I did not know exactly where it was, I will check.

JosePizarro3 (approved):
Very nice. Thanks!

Issue #170, opened 2023-10-24
Broken parsing of 0D systems in Crystal


ndaelman-hu:
This ticket was opened to follow up with a user-reported bug. This is the description from the e-mail:

> I have uploaded to Nomad a set of calculations for different polystyrene structures but in 3 out of 5 entries the processing part has failed. It seems that there is a problem when my structure was defined as an **isolated molecule (0D system)**. The calculations are done with **Crystal17** code. Have you encountered a similar issue with Crystal? Will it be possible to make the data public and accessible to everyone if the processing part has failed? The dataset name for the upload is **CHARISMA-PST-2023**.

ndaelman-hu:
Tracing down the error (see log below), the issue stems from missing cell information (`lattice` and `lattice_vectors`) at the height of `File "/usr/local/lib/python3.9/site-packages/electronicparsers/crystal/parser.py", line 658`.
When run in 0D mode, it appears that Crystal does not use/print out this cell data (under `LATTICE PARAMETERS  (ANGSTROMS AND DEGREES) - PRIMITIVE CELL`). There's nothing in the parser to handle this contingency.

@ladinesa I guess that I'll have to go through the entire parser and add checks to deal with missing lattice data. Do you see any alternatives, e.g. finding an appropriate default? On that last thought, do we now have the `np.infinity` property? Still, this might equally give issues down the line.

Error log:
```
Traceback (most recent call last):
File "/usr/local/lib/python3.9/site-packages/nomad/processing/data.py", line 1203, in parsing parser.parse(self.mainfile_file.os_path, self._parser_results, logger=logger, **kwargs)
File "/usr/local/lib/python3.9/site-packages/nomad/parsing/parser.py", line 398, in parse self.mainfile_parser.parse(mainfile, archive, logger)
File "/usr/local/lib/python3.9/site-packages/electronicparsers/crystal/parser.py", line 658, in parse cart_pos, atomic_numbers, atom_labels, lattice_vectors = to_system(
File "/usr/local/lib/python3.9/site-packages/electronicparsers/crystal/parser.py", line 1044, in to_system cart_pos = atomutils.to_cartesian(scaled_pos, lattice_vectors)
File "/usr/local/lib/python3.9/site-packages/nomad/atomutils.py", line 204, in to_cartesian cartesian_positions = np.dot(positions, complete_cell(cell))
[...]

PR #177, opened 2023-11-02
170 broken parsing of 0d systems in crystal

ndaelman-hu:
I fixed the lattice parameter assignment causing the bug in issue #170.
It now uses the `dimensionality` to assign `cart_pos`, rather than the system type.
I also cleaned up `to_system`, which assigned the Cartesian positions.

I slightly extended the tests, but I can see to also include an actual sample of the original bug report.
All examples from the bug report now pass, see figure.

Screenshot from 2023-11-02 23-36-52

### DIFF
--- a/electronicparsers/crystal/parser.py
+++ b/electronicparsers/crystal/parser.py
@@ -615,7 +615,6 @@ class CrystalParser:
         # System. There are several alternative sources for this information
         # depending on the run type.
         system = run.m_create(System)
-        material_type = out["material_type"]
         system_edited = out["system_edited"]
         labels_positions = out["labels_positions"]
         lattice_vectors_restart = out["lattice_vectors_restart"]
--- a/electronicparsers/crystal/parser.py
+++ b/electronicparsers/crystal/parser.py
@@ -197,6 +197,13 @@ class CrystalParser:
                             repeats=False,
                         ),
                         Quantity(
+                            "lattice_positions_raw",
+                            fr'AT\.IRR\.\s+AT\s+AT\.N\.\s+X\s+Y\s+Z\s*{br}' +\
+                            fr'((?:\s+{integer}\s+{integer}\s+{integer}\s+{flt}\s+{flt}\s+{flt}{br})+)',
+                            shape=(-1, 6),
+                            dtype=str,
+                        ),
+                        Quantity(
                             "labels_positions",
                             fr'\s+ATOM\s+X(?:/A|\(ANGSTROM\))\s+Y(?:/B|\(ANGSTROM\))\s+Z(?:/C|\(ANGSTROM\))\s*{br}' +\
                             re.escape(' *******************************************************************************') +\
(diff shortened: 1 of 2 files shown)


ndaelman-hu:
@ladinesa Let me know if you prefer that Lauri reviews, since he originally wrote the Crystal parser.

ndaelman-hu:
Note, the conversion in `to_system` seems somewhat superfluous, even after simplification.
Essentially, all coordinates are written as scaled wrt to the cell dimensions. In lower dimensionalities, the excluded axes are written in Cartesian coordinates. It's the scaled positions that are converted to Cartesian coordinates.

However, the first print-out of the geometric structure is fully Cartesian. You can find it under the line
```AT.IRR.  AT  AT.N.          X                   Y                   Z```
I could just extract the starting structure as such, without the need for converting. From what I can tell, we still need the current approach for updates after ionic steps.

**Comment by ndaelman-hu** [comment]
On the same thought, should we consider **changing our communication of the PBC conditions?** Technically, the code converts any non-periodic axis to one with 500 Angstrom. It is pretty verbose about this, printing it out in the lattice parameters. However, when a 0D system is run, it no longer prints out the lattice...

File: electronicparsers/crystal/parser.py [hunk 1]
```diff
@@ -608,20 +615,18 @@ def parse(self, filepath, archive, logger):
...
         # depending on the run type.
         system = run.m_create(System)
-        material_type = out["material_type"]
         system_edited = out["system_edited"]
         labels_positions = out["labels_positions"]
         lattice_vectors_restart = out["lattice_vectors_restart"]
-        pbc = None if material_type == "MOLECULAR CALCULATION" else np.array([True, True, True])
+        dimensionality = out["dimensionality"]
+        if dimensionality == 0:
+            pbc = np.array([False] * 3)
```
ladinesa: minor styling, can also be np.zeros(3, dtype=bool) and np.ones(3, dtype=bool)

electronicparsers/crystal/parser.py
```diff
@@ -196,6 +196,13 @@ def parse_output(self, filepath):
                             dtype=np.float64,
                             repeats=False,
                         ),
+                        Quantity(
+                            "lattice_positions_raw",
+                            fr'AT\.IRR\.\s+AT\s+AT\.N\.\s+X\s+Y\s+Z\s*{br}' +\
+                            fr'((?:\s+{integer}\s+{integer}\s+{integer}\s+{flt}\s+{flt}\s+{flt}{br})+)',
+                            shape=(-1, 6),
```
ladinesa: first time I have seen shape -1, good to know.

ladinesa:
Looks good. thanks for the fix

electronicparsers/crystal/parser.py
ndaelman-hu: I forgot that I left this capturing group in... It didn't work, but I fixed it now. Still, it isn't used anywhere atm.

ndaelman-hu:
> Note, the conversion in to_system seems somewhat superfluous, even after simplification.
> Essentially, all coordinates are written as scaled wrt to the cell dimensions. In lower dimensionalities, the excluded axes are written in Cartesian coordinates. It's the scaled positions that are converted to Cartesian coordinates.
> 
> However, the first print-out of the geometric structure is fully Cartesian. You can find it under the line
> AT.IRR.  AT  AT.N.          X                   Y                   Z
> I could just extract the starting structure as such, without the need for converting. From what I can tell, we still need the current approach for updates after ionic steps.

So, the problem lies with line 629:

```python
        # If any geometry edits (supercells, substitutions, dispplacements,
        # deformations, nanotube construction, etc.) are done on top of the
        # original system, they override the original system.
        if system_edited is not None:
            if system_edited["labels_positions_nanotube"] is not None:  # pylint: disable=E1136
                labels_positions = system_edited["labels_positions_nanotube"]  # pylint: disable=E1136
            else:
                labels_positions = system_edited["labels_positions"]  # pylint: disable=E1136
            # TODO adjust re pattern for other formats e.g. with R(ANGS)
            if labels_positions is not None:
                atomic_numbers = labels_positions[:, 2]  # pylint: disable=E1136
                atom_labels = labels_positions[:, 3]  # pylint: disable=E1136
                atom_pos = labels_positions[:, 4:7]  # pylint: disable=E1136
            if system_edited["lattice_parameters"] is not None:  # pylint: disable=E1136
                lattice = system_edited["lattice_parameters"]  # pylint: disable=E1136
   ```

Here, the format may be changed to include a mixture of scaled and Cartesian coordinates.

ndaelman-hu:
> Looks good. thanks for the fix

What do I do with `labels_positions_raw`?

ladinesa:
> > Looks good. thanks for the fix
> 
> What do I do with `labels_positions_raw`?

Not sure what this is for, it is the starting structure? if I understand the issue right, probably you can set the argument to to_system for dimensionality to 0 for labels_positions_raw.

ndaelman-hu:
> > > Looks good. thanks for the fix
> > 
> > 
> > What do I do with `labels_positions_raw`?
> 
> Not sure what this is for, it is the starting structure? if I understand the issue right, probably you can set the argument to to_system for dimensionality to 0 for labels_positions_raw.

Indeed, `labels_positions` and `labels_positions_raw` both extract the starting structure.
`labels_positions_raw` already has the right coordinates (all Cartesian) and can be stored without the need for `to_system`.

`lattice_positions` (and the subtypes) extract updates to the initial geometry (e.g. optimization, frequency, etc.).
Just as `labels_positions`, they all follow a mixed coordinate system (scaled for axes with PBC, and Cartesian for the rest).
These do require `to_system`.

So you see that we have 2 options on how to deal with this.

ladinesa:
> > > > Looks good. thanks for the fix
> > > 
> > > 
> > > What do I do with `labels_positions_raw`?
> > 
> > 
> > Not sure what this is for, it is the starting structure? if I understand the issue right, probably you can set the argument to to_system for dimensionality to 0 for labels_positions_raw.
> 
> Indeed, `labels_positions` and `labels_positions_raw` both extract the starting structure. `labels_positions_raw` already has the right coordinates (all Cartesian) and can be stored without the need for `to_system`.
> 
> `lattice_positions` (and the subtypes) extract updates to the initial geometry (e.g. optimization, frequency, etc.). Just as `labels_positions`, they all follow a mixed coordinate system (scaled for axes with PBC, and Cartesian for the rest). These do require `to_system`.
> 
> So you see that we have 2 options on how to deal with this.

Just maybe use labels_positions. Leave the defintion for labels_positions_raw for now.

PR #179, opened 2023-11-14
chore: use a separate action for ruff linting

blueraft:
Use ruff action instead of running ruff in the build pipeline.


ladinesa:
what is the advantage of this?

blueraft:
Ruff doesn't require any python dependencies, so it's possible to run it independently without waiting for python installation. This allows us to run the linting check in parallel to the tests, the linting check will run even if the python building step fails for whatever reason.

ladinesa (approved):

PR #181, opened 2023-11-15
Core hole vasp

ndaelman-hu:
Identification of the core-hole (location + orbital level) with appropriate references to `run.system.atoms_group` so topology view is leveraged. This will be further extended to FHIaims and ATOMPAW in separate PRs.

### DIFF
--- a/electronicparsers/openmx/parser.py
+++ b/electronicparsers/openmx/parser.py
@@ -16,6 +16,7 @@
 # limitations under the License.
 #
 
+from typing import Optional
 import numpy as np
 import re
 import io
@@ -31,7 +32,7 @@ from nomad.datamodel.metainfo.simulation.system import (
     System, Atoms
 )
 from nomad.datamodel.metainfo.simulation.method import (
-    Method, BasisSet, DFT, XCFunctional, Functional, Electronic, Smearing, Scf,
+    CoreHole, Method, BasisSet, DFT, Pseudopotential, XCFunctional, Functional, Electronic, Smearing, Scf,
     BasisSetContainer,
 )
 from nomad.parsing.file_parser import TextParser, Quantity
@@ -243,6 +244,51 @@ def parse_temperature_profile(data, thermo_type):
 class OpenmxParser:
     def __init__(self):
         pass
+
+    def parse_species(self, definition: str) -> tuple[Pseudopotential, Optional[CoreHole]]:
+        '''
+        Extract `Pseudopotential` and `CoreHole` (if present) from the atomic species definition.
+        An explanation of the format can be found at https://www.openmx-square.org/openmx_man3.9/node32.html
+        For an overview of all conventional potentials and partial atomic orbitals, see https://www.openmx-square.org/vps_pao2019/
+        and https://www.openmx-square.org/vps_pao_core2019/ for core-holes.
+        '''
+        definitions = definition.strip().split()
+        element = '[A-Z][a-z]?'
+        l_quantum = '[spdf]'
+        basis_set_orbital = f'{l_quantum}\d'
+        _remove_extension = lambda x: re.sub(r'[\.(pao, vps)]$', '', x)
+        _extract_method = lambda x: re.match(r'_([A-Z]+19)', x)
+        _extract_orbital = lambda x: re.match(r'_(\d)([spdf])$', x)
+        _extract_elem_cutoff = lambda x: re.match(rf'^({element})([\d\.]+)[_-]', x)
+        _extract_basis_set = lambda x: re.match(rf'-({basis_set_orbital}+)$', x)
+        _extract_lmax = lambda x: re.match(rf'({l_quantum})\d$', x)
+
+        definitions[1] = _remove_extension(definitions[1])
+        definitions[2] = _remove_extension(definitions[2])
+
+        # evaluate pseudopotential
+        pseudopotential, core_hole = Pseudopotential(), None  # TODO: add basis set
+        pseudopotential.name = f'{definitions[1]} {definitions[2]}'
+        pseudopotential.xc_functional_name = _extract_method(definitions[1]).group(1)
+        pseudopotential.cutoff = float(_extract_elem_cutoff(definitions[2]).group(2))
+        pseudopotential.type = 'US MBK'
+        pseudopotential.norm_conserving = True
+        pseudopotential.l_max = int(_extract_lmax(_extract_basis_set(definitions[2]).group(1)).group(1))
+
+        # evaluate core_hole
+        nq, lq = _extract_orbital(pseudopotential.name).groups()
+        if any(nq, lq):
+            l_mapping = dict(zip('spdf', range(4)))
+            core_hole = CoreHole(
+                n_quantum_number = int(nq),
+                occupation = 0.,
+            )
+            try:
+                core_hole.l_quantum_number = l_mapping[lq]
+            except KeyError:
+                raise ValueError('Unknown orbital type: {}'.format(lq))
+
+        return pseudopotential, core_hole
 
     def parse_workflow(self):
         md_type = mainfile_parser.get('MD.Type')
--- a/electronicparsers/openmx/parser.py
+++ b/electronicparsers/openmx/parser.py
@@ -51,6 +51,18 @@ This is parser for OpenMX DFT code.
 
 # A = (1 * units.angstrom).to_base_units().magnitude
 
+def _j_mapping() -> dict[tuple[int, int], tuple[float, float]]:
+    '''Reproduce table 12 given in https://www.openmx-square.org/openmx_man3.9/node192.html'''
+    mapping: dict[tuple[int, int], tuple[float, float]] = {}
+    for ll in range(4):
+        counter = 0
+        for jj in (.5, -.5):
+            j = ll + jj
+            for mj in range(-j, j + 1):
+                counter += 1
+                mapping[(ll, counter)] = (j, mj)
+    return mapping
+
 element = '[A-Z][a-z]?'
 xc_functional_dictionary = {
     'GGA-PBE': ['GGA_C_PBE', 'GGA_X_PBE'],
@@ -344,7 +356,10 @@ class OpenmxParser:
                 spinpol = mainfile_parser.get('scf.SpinPolarization', '').lower()
                 if spinpol == 'on':
                     core_hole.n_electrons_excited = 1.
-                    core_hole.ms_quantum_bool = not bool(int(core_hole_flags.results['core_hole'][2] / core_hole.degeneracy))
+                    core_hole.ms_quantum_bool = not bool(int(core_hole_flags.results['core_hole'][2] / core_hole.multiplicity))
+                elif spinpol == 'nc':
+                    core_hole.n_electrons_excited = 1.
+                    core_hole.j_quantum_number, core_hole.mj_quantum_number = _j_mapping()[core_hole.l_quantum_number, core_hole_flags.results['core_hole'][3]]
                 elif spinpol == 'off':
                     core_hole.n_electrons_excited = 2.  # TODO: verify
                     logger.warning('''
@@ -352,7 +367,7 @@ class OpenmxParser:
                     This is not recommended by the manual. For now assuming double electron (bosonic) excitation.
                     ''')
                 else:
-                    logger.warning('Could not set all quantum numbers: non-collinear calculations not yet supported')
+                    logger.warning('Spin-state not yet recognized')
             else:
                 core_hole.dscf_state = 'initial'  # this will be a hook in $\Delta$-SCF
                 core_hole.n_electrons_excited = 0.
--- a/electronicparsers/openmx/parser.py
+++ b/electronicparsers/openmx/parser.py
@@ -16,6 +16,7 @@
 # limitations under the License.
 #
 
+import logging
 from typing import Optional
 import numpy as np
 import re
@@ -32,7 +33,7 @@ from nomad.datamodel.metainfo.simulation.system import (
     System, Atoms
 )
 from nomad.datamodel.metainfo.simulation.method import (
-    CoreHole, Method, BasisSet, DFT, Pseudopotential, XCFunctional, Functional, Electronic, Smearing, Scf,
+    AtomParameters, CoreHole, Method, BasisSet, DFT, Pseudopotential, XCFunctional, Functional, Electronic, Smearing, Scf,
     BasisSetContainer,
 )
 from nomad.parsing.file_parser import TextParser, Quantity
@@ -49,6 +50,17 @@ This is parser for OpenMX DFT code.
 
 # A = (1 * units.angstrom).to_base_units().magnitude
 
+element = '[A-Z][a-z]?'
+xc_functional_dictionary = {
+    'GGA-PBE': ['GGA_C_PBE', 'GGA_X_PBE'],
+    'LDA': ['LDA_X', 'LDA_C_PZ'],
+    'LSDA-CA': ['LDA_X', 'LDA_C_PZ'],
+    'LSDA-PW': ['LDA_X', 'LDA_C_PW'],
+    None: ['LDA_X', 'LDA_C_PZ']
+}
+xc_functional_dictionary['PBE'] = xc_functional_dictionary['GGA-PBE']
+xc_functional_dictionary['CA'] = xc_functional_dictionary['LSDA-CA']
+
 scf_step_parser = TextParser(quantities=[
     Quantity('NormRD', r'NormRD=\s*([\d\.]+)', repeats=False),
     Quantity('Uele', r'Uele=\s*([-\d\.]+)', repeats=False)
@@ -60,8 +72,13 @@ md_step_parser = TextParser(quantities=[
 ])
 
 species_and_coordinates_parser = TextParser(quantities=[
-    Quantity('atom', r'\s*\d+\s*([A-Za-z]{1,2})\s*([-\d\.]+)\s+([-\d\.]+)\s+([-\d\.]+)\s+[\d\.]+\s*[\d\.]+\s*',
+    Quantity('atom', rf'\s*\d+\s*({element}\d*)\s*([-\d\.]+)\s+([-\d\.]+)\s+([-\d\.]+)\s+[\d\.]+\s*[\d\.]+\s*',
              repeats=True)
+])
+
+species_definition_parser = TextParser(quantities=[
+    Quantity('species', rf'({element}\d*)\s+([-\w\.]+)\s+(\w+)',
+             repeats=True),
 ])
 
 
@@ -88,6 +105,9 @@ mainfile_parser = TextParser(quantities=[
         'atoms', r'<Atoms.SpeciesAndCoordinates([\s\S]+)Atoms.SpeciesAndCoordinates>',
         sub_parser=species_and_coordinates_parser,
         repeats=False),
+    Quantity(
+        'species', r'<Definition\.of\.Atomic\.Species([\s\S]+)Definition\.of\.Atomic\.Species>',
+        sub_parser=species_definition_parser,),
     Quantity(
         'input_lattice_vectors', r'(?i)<Atoms.UnitVectors\s+((?:-?\d+\.\d+\s+)+)Atoms.UnitVectors>',
         repeats=False),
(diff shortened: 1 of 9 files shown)


ndaelman-hu:
Left to do:
- [x] Apply ICORELEVEL tag at the electron count level
- [x] Check linting
- [ ] Add core-hole test (with varying occupations or multiple sites?)
- [ ] Add documentation?

electronicparsers/vasp/parser.py
```diff
@@ -1366,7 +1371,58 @@ def parse_kpoints(self, section):
...
+                is_molecule=False,
+            )
+            for i in atom_ids
+        ]
+        atoms_group = AtomsGroup(
+            label='core-hole_' + elem_label,
+            type='core_hole',
+            atom_indices=atom_ids,
+            n_atoms=len(atom_ids),
+            atoms_group=sub_groups,
```
JosePizarro3: Yes, this is the wrong point I am mentioning in the Gitlab. I will do a more thorough review, but initially I'd say that `atoms_group` is an unecesserary repetition, as you are creating this extra layer.

You only need to store something like:
```
atoms_group = AtomsGroup(
   type='core_hole',
   label=elem_label,
   n_atoms=len(atom_ids),
   atom_indices=atom_ids,
)
```

Anything else (`label='core-hole_' + elem_label` and `atoms_group=sub_groups`) is the repetition I am mentioning that you don't need 🙂

JosePizarro3 (changes requested):

electronicparsers/vasp/parser.py
ndaelman-hu: > initially I'd say that atoms_group is an unecesserary repetition

Point taken. Let's keep this discussion over at Gitlab, and if we decide against this layer, I'll simply remove it.

ndaelman-hu: > Anything else (label='core-hole_' + elem_label and atoms_group=sub_groups) is the repetition I am mentioning that you don't need

Yeah, the current `label` was just a placeholder, so feel free to raise suggestions.
I applied your code, but this means that the term _core-hole_ isn't visible anymore at the 2nd visual layer (fig. 1).
If we go this way, we really do need a 3rd visual layer, just to denote the type. Else, this information should be in the title / label.

Something similar happens with Projection, where the word doesn't show up at all (fig. 2)... Granted, the calculation is labeled as a Wannier projection, so the user can likely guess what the sublayer means. Note that **this does not apply to a calculation with a core-hole**: it's just a regular `SinglePoint`, `GeometryOptimization`, etc.
I'd keep it that way.

Btw, the **all-caps is a nightmare for elemental labels**. In the examples, it's just TI or CU, respectively. I'd like to amend that together with Lauri.

Screenshot from 2023-11-17 20-15-39

Screenshot from 2023-11-17 20-17-41

electronicparsers/vasp/parser.py
```diff
@@ -1434,7 +1489,11 @@ def parse_method(self):
                     sec_hubb.double_counting_correction = self.hubbard_dc_corrections.get(parsed_file.get('LDAUTYPE'))
                     sec_hubb.x_vasp_projection_type = 'on-site'
             atom_counts[element[i]] += 1
-        sec_method.x_vasp_atom_kind_refs = sec_method.atom_parameters
+        sec_method.x_vasp_atom_kind_refs = sec_method.atom_parameters  # TODO: deprecate x_vasp_atom_kind_refs?
```
JosePizarro3: Yes, please. Deprecate it.

electronicparsers/vasp/parser.py
```diff
@@ -1547,7 +1610,7 @@ def parse_workflow(self):
                     workflow.method.convergence_tolerance_force_maximum = abs(tolerance) * ureg.eV / ureg.angstrom
         self.archive.workflow2 = workflow
 
-    def parse_configurations(self):
+    def parse_configurations(self, atoms_group: Union[AtomsGroup, None] = None):
```
JosePizarro3: Is this the correct typing? Shouldn't be `atoms_group: AtomsGroup = None`?

electronicparsers/vasp/parser.py
```diff
@@ -1834,8 +1904,11 @@ def parse(self, filepath, archive, logger):
         if self._calculation_type == 'gw':
             self.parse_gw()
         else:
-            self.parse_method()
+            method_dict = self.parse_method()
 
-        self.parse_configurations()
+        # Note: the logic for a proper topology might have to become more complex
+        self.parse_configurations(
```
JosePizarro3: It seems to me that `self.parse_configurations` will be more readable if using `method_dict` as an input, and then within the function you resolve `atoms_group`.

Just a small style detail.

JosePizarro3 (commented):
Very nice. Please, don't forget to change also `workflowparsers/quantum_espresso_xspectra` with the new `CoreHoleSpectra`.

I will keep up with the discussion about nesting AtomsGroup.

**Btw, can you share some examples and add testing?**

electronicparsers/vasp/parser.py
JosePizarro3: My main concern here is that: 3 layers is unnecessary, as we only need (1) the original crystal + (2) the specific `active_atoms` used and their environments. 

Now I see that the naming for `label` might be confusing. We can then go back to your case (`label=f'core_hole_{elem_id}'`) or define in gitlab a new structural type called `'active_atoms'` (or `['core_hole_atoms', 'projection_atoms']`. I'd personally prefer the first option, but maybe the second has its benefits

ondracka:
I can't comment much on the VASP parser changes, but rather on the general metainfo changes (doing this here due to no access to the upstream github). In general this would suit my needs, but at least for Wien2k for states with n>1 where a spin-orbit splitting happens, like 2p states, I can actually specify if I want the corehole for 2p1/2 or 2p3/2, and equivalently for d and f states. I could likely also specify up and down spin state with Wien2k (but I never did that).
I checked the VASP docs and there is seems this is not possible, but I would expect that maybe other all-electron codes can do this as well.
My another favorite code OpenMX also allows all kinds of core-holes (there is nice table at https://www.openmx-square.org/openmx_man3.9/node192.html) there it also depends whether one has collinear or non-collinear spin polarization. So maybe we need more quantum numbers...

ondracka:
One question is if we also want to somehow explicitly include somewhere the charge info at this point. Together with with the corehole usually also comes the question about where to put the extra charge. I.e., if the electron was removed from the cell (and a homogeneous background change was added so that the calculations behave in a sane way) but it is actually a charged cell (this is usually done for insulators) or if the electron was placed at the Fermi level (this would make sense for core-hole in metals). This is not related just to the core-hole and is relevant also for other methods (for charged-defect calculations in insulators, etc.). I believe VASP always adds the background charge automatically for example (and it will probably diverge for a true charged cell anyway).)

ndaelman-hu:
> I can't comment much on the VASP parser changes, but rather on the general metainfo changes (doing this here due to no access to the upstream github). In general this would suit my needs, but at least for Wien2k for states with n>1 where a spin-orbit splitting happens, like 2p states, I can actually specify if I want the corehole for 2p1/2 or 2p3/2, and equivalently for d and f states. I could likely also specify up and down spin state with Wien2k (but I never did that). I checked the VASP docs and there is seems this is not possible, but I would expect that maybe other all-electron codes can do this as well. My another favorite code OpenMX also allows all kinds of core-holes (there is nice table at https://www.openmx-square.org/openmx_man3.9/node192.html) there it also depends whether one has collinear or non-collinear spin polarization. So maybe we need more quantum numbers...

Thank you for pointing that out! I'll open up tickets for OpenMX then, and merge the schema after having working parsers for both parsers.

ndaelman-hu:
> One question is if we also want to somehow explicitly include somewhere the charge info at this point. Together with with the corehole usually also comes the question about where to put the extra charge. I.e., if the electron was removed from the cell (and a homogeneous background change was added so that the calculations behave in a sane way) but it is actually a charged cell (this is usually done for insulators) or if the electron was placed at the Fermi level (this would make sense for core-hole in metals). This is not related just to the core-hole and is relevant also for other methods (for charged-defect calculations in insulators, etc.). I believe VASP always adds the background charge automatically for example (and it will probably diverge for a true charged cell anyway).)

Very good point! We keep track of the no. (valence) electrons and overall system charge under `run.method.electronic`. We could just improve the normalizers here and make this information more visible in the Entry Overview. This would allow for finding core-holes with the electron exited to the Fermi level or the vacuum.

But maybe we want something **more explicit** when core-holes are being used? VASP also allows partial core-holes, as well as initial and final state approximations. @ondracka Are these also used in Wien2k or OpenMX? That would help me decide on their placement in the schema.

ondracka:
> Thank you for pointing that out! I'll open up tickets for OpenMX then, and merge the schema after having working parsers for both parsers.

Let me know if you need some test-cases. Here is a simple C 1s corehole in OpenMX: https://nomad-lab.eu/prod/v1/staging/gui/search/calculations/entry/id/zLqikxv-mfCrI3sC9EuXmqfEX04T but I can also produce something simpler and more suitable for eventual inclusion in parser unit tests (or also for the other core-holes than 1s). For Wien2k I have some N1s core-hole: https://nomad-lab.eu/prod/v1/staging/gui/search/entries/entry/id/Fbk7yucgh7CxMsp0RKdR263J_OIj , O1s https://nomad-lab.eu/prod/v1/staging/gui/search/entries/entry/id/1GH5Rvm-t2tAYPrL9UdtkShaas3s, or Ti2p3/2 https://nomad-lab.eu/prod/v1/staging/gui/search/entries/entry/id/s5W753xr30EulmEUzsu75HMtxZMh coreholes in NOMAD already.

electronicparsers/vasp/parser.py
ndaelman-hu: It can be simplified, yes, but not to your suggestion. The type `AtomsGroup` only includes the class, not `None` objects.

I will change it to `def parse_configurations(self, atoms_group: Optional[None] = None):`

electronicparsers/vasp/parser.py
ndaelman-hu: Sounds fine. I'll just add an explanation on the exact meaning and use of `method_dict`.

ondracka:
> Very good point! We keep track of the no. (valence) electrons and overall system charge under `run.method.electronic`. We could just improve the normalizers here and make this information more visible in the Entry Overview. This would allow for finding core-holes with the electron exited to the Fermi level or the vacuum.
> 
> But maybe we want something **more explicit** when core-holes are being used? VASP also allows partial core-holes, as well as initial and final state approximations. @ondracka Are these also used in Wien2k or OpenMX? That would help me decide on their placement in the schema.

Partial core-holes with arbitrary fraction are easy in Wien2k (and probably other LAPW all electron codes), however I'm not aware that this would be possible in OpenMX.
Regarding the "initial" states, i.e., I assume you are referring to what VASP would be doing with ICORELEVEL=1 this is what all-electron codes do by default (and at that point there is no core-hole at all actually, just VASP includes the core-levels in the calculations instead of using pseudopotential), so for all-electron codes this is just standard groundstate. For OpenMX "initial state" one actually needs specific pseudopotential and basis set for the atom with the core-hole that include the core-level in question, but no other setting are needed. So I think that for VASP run with ICORELEVEL=1 there would be no CoreHole section and the fact that the extra core electron are explicitly included would be encoded somehow in the pseudopotential section.
"Final state" is the true core-hole calculation.

electronicparsers/vasp/parser.py
JosePizarro3: That's a weird typing (Optional None to None?), but nvm, this is then deprecated if you pass `method_dict: dict = {}` 😄

ndaelman-hu: Oops, that's a typo. It has to be `def parse_configurations(self, atoms_group: Optional[AtomsGroup] = None)`, obviously.

ndaelman-hu:
> but at least for Wien2k for states with n>1 where a spin-orbit splitting happens, like 2p states, I can actually specify if I want the corehole for 2p1/2 or 2p3/2, and equivalently for d and f states. I could likely also specify up and down spin state with Wien2k (but I never did that).

This is supported now.

> My another favorite code OpenMX also allows all kinds of core-holes (there is nice table at https://www.openmx-square.org/openmx_man3.9/node192.html) there it also depends whether one has collinear or non-collinear spin polarization. So maybe we need more quantum numbers...

Let me look into this. Maybe support for non-collinear calculations requires its own PR. Feel free to make any suggestions.

ndaelman-hu:
> Let me know if you need some test-cases. Here is a simple C 1s corehole in OpenMX: https://nomad-lab.eu/prod/v1/staging/gui/search/calculations/entry/id/zLqikxv-mfCrI3sC9EuXmqfEX04T but I can also produce something simpler and more suitable for eventual inclusion in parser unit tests (or also for the other core-holes than 1s).

@ondracka It seems that OpenMX reads the pseudopotentials from a different folder. Since the files are not copied over to the calculation folder, I guess it's unlikely that people will upload them.

Personally, I prefer reading the original pseudopotential file, but I guess that here it's better to build out an interpreter of the filename. Do you agree? How reliable is this approach?

ndaelman-hu:
> Regarding the "initial" states, i.e., I assume you are referring to what VASP would be doing with ICORELEVEL=1

Indeed, this is the terminology they use in the [VASP Wiki](https://www.vasp.at/wiki/index.php/ICORELEVEL).

> this is what all-electron codes do by default (and at that point there is no core-hole at all actually, just VASP includes the core-levels in the calculations instead of using pseudopotential), so for all-electron codes this is just standard groundstate.

@ondracka I fully agree with your explanation of the methodology.
However, in most contexts, ICORELEVEL=1is employed with Koopmans theorem or Slater's Transition method to produce CLBEs. One such example (not with VASP) would be [here](https://pubs.aip.org/aip/jcp/article/158/9/094111/2881303/Slater-transition-methods-for-core-level-electron). An overview of the various approaches is given in the first 2 paragraphs on page 2.

So, I see 2 options:
1. I infer that anyone using I**CORELEVEL**=1, does so with the intention of computing a core-level. There are several ways to approximate this (Koopman, or any of the fractional occupations listed above). If I know which state is targeted, I could even produce all the numbers.
2. I leave it up to the user to specify if they used the orbitals for computing the core-hole, as well as the actual scheme they used. This would come closer to what you proposed.

Let me know what you think.

ondracka:
> Let me look into this. Maybe support for non-collinear calculations requires its own PR. Feel free to make any suggestions.

I think this is OK for now, lets revisit the non-collinear case when we actually have some use cases...

ondracka:
> Personally, I prefer reading the original pseudopotential file, but I guess that here it's better to build out an interpreter of the filename. Do you agree? How reliable is this approach?

I'm only aware of the one set of pseudos provided by the OpenMX developers (versioned by the releases), so it should be quite reliable (unless people really handcraft their own pseudos, but that should be not too common).

ondracka:
I went quickly through the actual code changes and while I was not reviewing the logic per se, the general idea how things are done now looks fine, so I'm happy with how this turned out in general.

ndaelman-hu:
@ladinesa @lauri-codes Design question: core-holes are kind of connected to various other sections (too much interconnectivity):
1. Each `run.method.CoreHole` has to reference an `run.system.AtomsGroup` (this also applies to other sections, such as `CoreHoleSpectra`). Atm, I set these manually in the parsers, but I get the feeling that this could be done by a normalizer. I guess one could loop over all `AtomParameters` which contain `CoreHole`, but it sounds suboptimal: it requires the atom indexes too... Moreover, they might be other sections that do not fall under `AtomsParameters`. I'd rather have a push-system that works for any `AtomsGroup` - Methods.subs_section pair. Any ideas?
2. The computation of `Method.charge` is contingent on `CoreHole`. Same remark as above.
3. `CoreHole` is in many cases specified in the same parsed blocks / files as `Pseudopotential`. I'm trying to package both together in each new implementation now.

lauri-codes:
I think the parser is ultimately responsible for setting all connections in place. A normalizer could very well be used to e.g. check the validity of the data or to add new derived data.

Note that it is of course fine to have a single unified **function** in e.g. `nomad` or in the `electronicparsers` package that is shared between all parsers and resolves all the references. I think this approach is often used in the `atomisticparsers` package, where several tasks are common for the parsers.

ndaelman-hu:
> I think the parser is ultimately responsible for setting all connections in place. A normalizer could very well be used to e.g. check the validity of the data or to add new derived data.
> 
> Note that it is of course fine to have a single unified **function** in e.g. `nomad` or in the `electronicparsers` package that is shared between all parsers and resolves all the references. I think this approach is often used in the `atomisticparsers` package, where several tasks are common for the parsers.

I get your point, in that the logical structure of the output may vary and this is where the parser has to jump in, of course.
However, the schema on our side will always be `AtomsGroup` pointing to `CoreHole`, where the latter may alter the overall `charge` (and is often set along with the `Pseudopotential`).

For anyone extending the parsers, this structure is far from obvious, nor is it clearly communicated anywhere.
My first instinct, for example, was to extend `AtomsGroup`, rather than `Method`. I think we have to communicate these kinds of designs better to other (external) devs, since we're headed in the direction of a modular platform.

Maybe a _normalizer_ is not appropriate here. Another option that come to mind is a `FactoryMethod`, which delivers (maybe even sets) a `AtomsGroup` - `CoreHole` pair. If we go this route, where would this best be added? In `util.py`?

lauri-codes:
My first suggestion would be to include the method in the `electronicparsers` repo if it is only relevant for these parsers. But if the implementation should be shared across different parser repos, then `util.py` is an option.

ndaelman-hu:
@ondracka From the [OpenMX documentation](https://www.openmx-square.org/openmx_man3.9/node192.html), I'm understanding that CLBEs are computed using $\Delta$-SCF between the _initial_ (containing the corrected pseudopotential) and _final_ state (corrected pseudopotential + electron removed). Is this correct?

I'm asking, since the terminology is confusing with [VASP's](https://www.vasp.at/wiki/index.php/ICORELEVEL), where _initial-state approximation_ (`icorelevel = 1`) is a single SCF with Koopman. There, the $\Delta$-SCF is between a regular groundstate (same pseudopotential) and a _final-state approximation_ (`icorelevel = 2`).

ondracka:
> @ondracka From the [OpenMX documentation](https://www.openmx-square.org/openmx_man3.9/node192.html), I'm understanding that CLBEs are computed using Δ-SCF between the _initial_ (containing the corrected pseudopotential) and _final_ state (corrected pseudopotential + electron removed). Is this correct?

That is my understanding as well.

ndaelman-hu:
> > @ondracka From the [OpenMX documentation](https://www.openmx-square.org/openmx_man3.9/node192.html), I'm understanding that CLBEs are computed using Δ-SCF between the _initial_ (containing the corrected pseudopotential) and _final_ state (corrected pseudopotential + electron removed). Is this correct?
> 
> That is my understanding as well.

Okay, great! I'm close to finishing with OpenMX too, will request you to review then.

JosePizarro3:
Btw, is something included in the TBStudio parser regarding the active atom?

ndaelman-hu:
> Btw, is something included in the TBStudio parser regarding the active atom?

Hey, thx for mentioning TBStudio. I had to rebase first, now I can see that parser too.
Atm, it does not assign any `atoms_group`s, so it won't break with any of my changes.
I will add support when extending `core_hole` to atomic orbitals in general. I'll open an issue this evening.

electronicparsers/openmx/parser.py
```diff
@@ -244,6 +286,91 @@ class OpenmxParser:
     def __init__(self):
         pass
 
+    @property
```
ladinesa: why is this a  property?

electronicparsers/openmx/parser.py
```diff
@@ -244,6 +286,91 @@ class OpenmxParser:
...
+        extract_core_hole = lambda x: re.search(rf'_(\d)({l_quantum})_CH', x)
+        extract_elem_cutoff = lambda x: re.match(rf'({element})([\d\.]+)[_-]', x)
+        extract_lmax = lambda x: re.search(rf'({l_quantum})\d$', x)
+
+        definitions = deepcopy(definitions)
+        try:
+            definitions[1] = remove_extension(definitions[1])
+            definitions[2] = remove_extension(definitions[2])
+        except IndexError:
+            logger.error(f'Species definition must be of length 3: {definitions}')
```
ladinesa: should you return here?

electronicparsers/openmx/parser.py
```diff
@@ -244,6 +286,91 @@ class OpenmxParser:
...
+        if quantum_nums_flag:
+            quantum_nums = quantum_nums_flag.groups()
+            try:
+                core_hole = CoreHole(
+                    n_quantum_number=int(quantum_nums[0]),
+                    l_quantum_number=CoreHole.l_quantum_numbers[quantum_nums[1]],
+                    n_electrons_excited=1,
+                )
+            except KeyError:
+                logger.error(f'Unknown l-quantum symbol: {quantum_nums[1]}')
```
ladinesa: should you return here?

electronicparsers/openmx/parser.py
```diff
@@ -414,7 +545,7 @@ def parse(self, mainfile: str, archive: EntryArchive, logger):
 
         self.parse_workflow()
 
-        self.parse_method()
+        self.parse_method(logger)
```
ladinesa: no self.logger?

electronicparsers/openmx/parser.py
```diff
@@ -471,4 +602,23 @@ def parse(self, mainfile: str, archive: EntryArchive, logger):
                         if time is not None:
                             sec_calc.time = time * units.fs
 
+        # set core-hole atoms groups
+        sec_system = self.archive.run[-1].system[-1]
```
ladinesa: error handling here

electronicparsers/vasp/parser.py
```diff
@@ -1497,6 +1531,26 @@ def parse_method(self):
...
+        neutral_count = 0.
+        for param, n_atoms in dict(zip(sec_method.atom_parameters, self.parser.atom_info['atomtypes']['atomspertype'])).items():
+            # correct based on core-holes
+            # since ZVAL information is centrally reported, it's all or nothing
+            try:
+                species_electrons = param.n_electrons
+            except AttributeError:
+                break
+            try:
+                species_electrons -= param.core_hole.n_electrons_excited
```
ladinesa, line 1545: what is the initial value for species_electrons

electronicparsers/vasp/parser.py
```diff
@@ -1575,6 +1629,20 @@ def parse_system(n_calc):
             if nose is not None:
                 sec_system.x_vasp_nose_thermostat = nose
 
+            # Check for core-holes
+            ## note that this check should be added to `parse_core_hole` if other kinds of atom_parameters are set
+            if self.parser.incar.get('ICORELEVEL', 0) == 2:
+                core_hole, core_hole_group, corehole_id = self.parse_core_hole()
+                self.archive.run[-1].method[-1].atom_parameters[corehole_id].core_hole = core_hole
```
ladinesa: no need for error handling here?

ladinesa (commented):
Some miinor suggestions.

electronicparsers/openmx/parser.py
ndaelman-hu: Simply easier access: essentially this is a dynamically generated dictionary
It's accessed 2x later on

electronicparsers/openmx/parser.py
ndaelman-hu: Is it not passed as an argument of `parse`?

electronicparsers/openmx/parser.py
ladinesa: why not simply have a get function. a property should depend on a state somehow.

electronicparsers/openmx/parser.py
ladinesa: what i meant was you do not need to pass as arg to parse_method if it is class attr.

electronicparsers/vasp/parser.py
ndaelman-hu: Isn't it `param.n_electrons`? Else the loop breaks.

ladinesa: i see

electronicparsers/vasp/parser.py
ndaelman-hu: I held off on resolving this discussions due to changes at the time.
No arguments are being passed to `parse_configurations`, so it is essentially taken care off.

Just a small point of feedback. While your suggestion definitely improves readability, it comes with the downside of coupling the input of `parse_configurations` to the output from `parse_method`. That coupling is too strong, and therefore a fatal design flaw.
In this case, it would be better to keep this format (closer to functional programming), OR have `parse_configurations` collect the various output packages of other methods `parse_...`.
In that case it's still coupled, but less. The final decision depends on how likely those methods are to be removed.

electronicparsers/openmx/parser.py
ndaelman-hu: I'm sorry, I'm not seeing the distinction: is the `atom_index_dict` a state of `self.archive`?
I can also change this to a `cached_property` decorator, but that might be overkill for 2-time use

ndaelman-hu: Sorry, it's actually accessed in a loop, so caching might be advantageous

electronicparsers/vasp/parser.py
ndaelman-hu: can I resolve?

electronicparsers/openmx/parser.py
ladinesa: but everytime you access this, it evaluates it. you can have self._atom_index_dict that can be reset if some dependency changes.

electronicparsers/openmx/parser.py
ndaelman-hu: Is it a class attribute? I'm only seeing it passed as an argument to `OpenmxParser.parse()`.
Moreover, wouldn't the alternative be more of an object attribute instead?

ladinesa: Yes that was I was asking if it is a class attr, if not maybe add it under parse.

electronicparsers/vasp/parser.py
```diff
@@ -1575,6 +1629,20 @@ def parse_system(n_calc):
...
+
+                if sec_system.atoms_group is None:
+                    sec_system.atoms_group = []
```
ndaelman-hu, line 1643: @ladinesa I guess this is redundant then, thanks to the instantiation process

ladinesa: yes it is. I think this is still a remnant of the old infra where the default for repeating sections is None.

electronicparsers/openmx/parser.py
ndaelman-hu: > but everytime you access this, it evaluates it

as is, yes, but not if I cache it

electronicparsers/openmx/parser.py
ndaelman-hu: I was following the example of `parse_structure`, but this is a free-standing method.
We might want to stratify the interface.

ladinesa: yes this is already implemented in the abstract class for parsers that I am developing in https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1538

ndaelman-hu:
> Very nice. Please, don't forget to change also `workflowparsers/quantum_espresso_xspectra` with the new `CoreHoleSpectra`.
> 
> I will keep up with the discussion about nesting AtomsGroup.
> 
> **Btw, can you share some examples and add testing?**

@JosePizarro3 Can you pls lift the change request in your review. That applied to an old version.
Your suggestions have been incorporated.

JosePizarro3 (approved):

PR #183, opened 2023-11-21
Fixing geomopt steps in CP2K

JosePizarro3:
@ladinesa

I was fixing the CP2K parser for the PDOS from Andrea.

It seems there is some specific SinglePoint cases that sill do not work. I attach here a smaller example from Andrea's dataset:

[singlepoints.zip](https://github.com/nomad-coe/electronic-parsers/files/13426784/singlepoints.zip)

It seems the regex for `quickstep.get('...')` is not working. Sorry, but because I don't have time before going on vacations, do you mind taking a look and fix it? Once you fix it, feel free to merge this and update the ref to the submodule.

### DIFF
--- a/electronicparsers/cp2k/parser.py
+++ b/electronicparsers/cp2k/parser.py
@@ -484,7 +484,7 @@ class CP2KOutParser(TextParser):
         geometry_optimization_quantities = [
             Quantity(
                 'method',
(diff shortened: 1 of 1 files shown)


ladinesa:
ok i will have look.

andreaa93:
let me know if I can give any help or test something

PR #188, opened 2023-11-28
Remove workflow schema from nomad

ladinesa:
Simulation workflow schema are moved to a plug in https://github.com/nomad-coe/nomad-schema-plugin-simulation-workflow. References to the old schema are removed.


ladinesa:
@ndaelman-hu can I simply merge this? I would like to merge the nomad mr soonish. thanks

ndaelman-hu (commented):
Pls double-check `./electronicparsers/charmm/metainfo/charmm.py`, line 1335: `class MolecularDynamics(simulation.workflow.MolecularDynamics):`.

A cursory search did not indicate any other occurrences.
I have the impression that you used `sed` to replace the pattern, but only applied it to `nomad.datamodel.metainfo.simulation.workflow`.
There's also the option to use `Rename Symbol` in VS, which tends to have a better coverage.

Overall, I like this change. It's much clearer and will help with migrating to base classes. Thx!

ndaelman-hu:
@JosePizarro3 it's good that you know this migration happened.

ladinesa:
> Pls double-check `./electronicparsers/charmm/metainfo/charmm.py`, line 1335: `class MolecularDynamics(simulation.workflow.MolecularDynamics):`.
> 
> A cursory search did not indicate any other occurrences. I have the impression that you used `sed` to replace the pattern, but only applied it to `nomad.datamodel.metainfo.simulation.workflow`. There's also the option to use `Rename Symbol` in VS, which tends to have a better coverage.
> 
> Overall, I like this change. It's much clearer and will help with migrating to base classes. Thx!

nice catch thanks

PR #195, opened 2024-02-05
Fix CP2K restart file

JosePizarro3:
See [#1879](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/issues/1879)

### DIFF
--- a/electronicparsers/cp2k/parser.py
+++ b/electronicparsers/cp2k/parser.py
@@ -927,7 +927,7 @@ class CP2KParser:
                 filename = self._normalize_filename(filename)
                 traj_format = self.inp_parser.get('MOTION/PRINT/TRAJECTORY/FORMAT', 'XYZ').strip()
                 traj_format = self._file_extension_map.get(traj_format, 'xyz')
-                filename = '%s-pos-1.%s' % (filename, traj_format)
+                filename = f'{filename}-pos-1.{traj_format}'
                 frequency = 1
 
             self.traj_parser.mainfile = os.path.join(self.maindir, filename)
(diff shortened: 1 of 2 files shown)


electronicparsers/cp2k/parser.py
```diff
@@ -1169,8 +1172,16 @@ def parse(name, data, section):
 
         input_files = get_files(input_filename, self.filepath, self.mainfile, deep=False)
         if not input_files:
-            self.logger.warning(f'Input file not found.')
-            return
+            self.logger.warning(
+                'Input .inp or .restart files not found. We will attempt finding the input '
+                'file if CP2K has appended -1 to the .restart file name.'
```
andreaa93: the -1 part is not clear, I think. 

We would need not to use the input_filename + "-1" but the project_filename fetched from the output file. 

What about a warning like this?

f" Input .inp not found. We will attempt finding the restart file with following sintax 'project_name'-1.restart: {project_name}-1.restart"

electronicparsers/cp2k/parser.py
```diff
@@ -1169,8 +1172,16 @@ def parse(name, data, section):
...
+            )
+            # Patch to check if the input is .restart and CP2K appended a -1 after the name
+            if '.restart' in input_filename:
+                appended_filename = input_filename.split('.restart')[0] + '-1.restart'
```
andreaa93: here, you should parse the project_name instead using again the input_filename

andreaa93 (changes requested):
I think project_name variable should be fetched from the very first lines of the output file, and we should try to see whether such a restart file is found.

In summary the possible filenames one can find are:
* input_filename
* {project_name}-1.restart

There is not the case in which a file such {input_filename}-1.restart is generated by cp2k

JosePizarro3:
Ok, I am extracting now `project_name = self.settings.get('global', {}).get('project_name')` and using it to append `f'{project_name}-1.restart'`.

Feel free to resolve the previous comments.

andreaa93 (approved):
All good!

PR #197, opened 2024-02-14
Add molpro electronic

ndaelman-hu:
Add Molpro parser as submodule


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "electronicparsers/molpro"]
+        path = electronicelectronicparsers/molpro
```
ladinesa: check path

PR #200, opened 2024-02-20
198 cp2k fatal error getting atomic numbers

### LINKED ISSUE #198: CP2K fatal error getting atomic numbers
When processing files like `supercell.xyz` in the [following entry](https://nomad-lab.eu/prod/v1/gui/dataset/id/BoTQoJACSY2YvRA-2WXyow/entry/id/WvqoStnQ1jaSFl3jbrYe6uxo1ABn), the parser fails in `get_atomic_number`, not finding the correct atomic number.

ndaelman-hu:
The error was with the header matching (newer versions) of an output section.
Still, we shouldn't need that section to produce atomic numbers from kinds.

closes #198


andreaa93 (commented):
I think it's fine, 

a question on tests: can you use in each test different versions of .out files to test on?

This can be done here or in another MR

ndaelman-hu:
> I think it's fine,
> 
> a question on tests: can you use in each test different versions of .out files to test on?
> 
> This can be done here or in another MR

I asked the user who reported this bug if I can use a file of theirs. Will see if I can get it in before merging.

andreaa93:
> > I think it's fine,
> > a question on tests: can you use in each test different versions of .out files to test on?
> > This can be done here or in another MR
> 
> I asked the user who reported this bug if I can use a file of theirs. Will see if I can get it in before merging.

I see, yeah that would be the best option I believe. And then, you add that file to be fed into tests. It's likely that some other test may fail with different versions of files so it's a nice way of making the parser more robust.

ladinesa:
please rebase so I can see the cp2k parser fixes. thanks.

electronicparsers/cp2k/parser.py
```diff
@@ -1147,13 +1148,16 @@ def _normalize_filename(self, filename):
...
-        )
-        return atomic_numbers.get(element, 0)
+    def get_atomic_number(self, element: str) -> Optional[int]:
+        """Convert the element symbol to its corresponding atomic number.
+
+        Dependencies:
+        - ase.data.atomic_numbers
+        """  # TODO: migrate responsilbity to section normalizer
+
+        if match := re.match(r"([A-Z][a-z]?)", element):
```
ladinesa, line 1158: no need to match,  why not use the old way and make this alternative: atomic_numbers.get(element, ase.data.atomic_numbers.get(element)). Is there a case where a non-standard symbol was used but the atom number was defined?

ndaelman-hu: @aalbino2 said that the element symbols may have a numerical suffix (see issue #198 ).
I obviously only want their name.

ladinesa: This will simply return none if not in ase.data matching will only introduce an unnecessary overhead i think

ndaelman-hu: No: `ase` does not recognize the suffixes. I just tried it out:

```python
ase.data.atomic_numbers.get("C1", None)

ase.data.atomic_numbers.get("C", None)
6
```

ladinesa: Yes exactly, it will return None if it contains any artifact aside from the element name, which you are also matching. This means we do not have to do regex matching beforehand

ndaelman-hu: Huh? I'm not following you. I'm only matching the element name:

``` python
re.match(r"([A-Z][a-z]?)", "C1")
<re.Match object; span=(0, 1), match='C'>
```

ladinesa: And this is not necessary.  you can directly pass element in ase.data.atomic_numbers.get(element, None)

ndaelman-hu: So, as we discussed in local chat, `re.match` should stay, though it could be potentially replaced by `ase.symbols.symbols2numbers`.
The latter produces a list of identical integers, which then need extraction, so I think that `re.match` is briefer and semantically clearer.

PR #202, opened 2024-02-21
Small QE fixes

ondracka:
- fix forces parsing (was broken in all QE versions since 2016, so actually 100% of NOMAD entries, just the ancient tests here worked)
- fix smearing parsing
- add n_spin_channels parsing
- two new testcases from the latest 7.3 version, mostly just used to test the new features, but I can also add more tests of the old functionality there if needed.

CC @ladinesa  @ndaelman-hu


electronicparsers/quantumespresso/parser.py
```diff
@@ -3248,6 +3249,15 @@ def parse_method(self, run):
...
         sec_method.electronic = sec_electronic
 
+        starting_magnetization = run.get_header("starting_magnetization")
+        spin_orbit_mode = run.get_header("spin_orbit_mode")
+        if starting_magnetization is not None:
+            sec_electronic.n_spin_channels = 2
+        elif spin_orbit_mode is not None:
+            sec_electronic.n_spin_channels = 4
+        else:
+            sec_electronic.n_spin_channels = 1
```
ndaelman-hu: What about data artifacts stemming from `spin_orbit_mode`, e.g. DOS, band structure?
Will those support the new shape?

electronicparsers/quantumespresso/parser.py
```diff
@@ -2784,8 +2784,9 @@ def __init__(self):
             "Marzari-Vanderbilt": "marzari-vanderbilt",
             "Methfessel-Paxton": "methfessel-paxton",
             "gaussian": "gaussian",
+            "Gaussian": "gaussian",
             "Fermi-Dirac": "fermi",
-            "tetrahedron": "tetrahedra",
+            "(tetrahedron": "tetrahedra",
```
ndaelman-hu: This looks like it's accommodating a faulty regex...

tests/test_quantumespressoparser.py
```diff
@@ -20,6 +20,7 @@
 import numpy as np
 
 from nomad.datamodel import EntryArchive
+from nomad.units import ureg as units
```
ndaelman-hu: Most of the code base uses `ureg` directly. No need to rename it.

ndaelman-hu (commented):
Thank you for the quick patches and tests! Just some minor questions / comments.

When we add more schema for non-collinear calculations in the future, I'd like to have your feedback too.

ndaelman-hu:
Btw, Pavel, is there any specific reason that you prefer to fork the branch?
If it were a branch on this project, it'd be easier for me to check out locally.

ondracka:
> Btw, Pavel, is there any specific reason that you prefer to fork the branch? If it were a branch on this project, it'd be easier for me to check out locally.

Just habits from other opensource projects I contribute to. I don't usually have write access to the other projects. Also when you have more people working on a project, every git fetch could result in hundreds of new branches and will clutter the project a lot, so working in a fork is kind of cleaner in this regards, but I could push to the main repo in the future, no problem...

ndaelman-hu:
> > Btw, Pavel, is there any specific reason that you prefer to fork the branch? If it were a branch on this project, it'd be easier for me to check out locally.
> 
> Just habits from other opensource projects I contribute to. I don't usually have write access to the other projects. Also when you have more people working on a project, every git fetch could result in hundreds of new branches and will clutter the project a lot, so working in a fork is kind of cleaner in this regards, but I could push to the main repo in the future, no problem...

I see. In the future, each code will have its own project, so the `git fetch` shouldn't trigger any massive download :p

electronicparsers/quantumespresso/parser.py
ondracka: I'm in no way expert here, I wanted to add `n_spin_channels` just to parse whether something is spin-polarized or not (and added the non-collinear part just for completeness). However I believe that with stuff like spin-orbit-coupling (SOC) you have just a single DOS, eigenvalues, etc... The way I understand, when you have collinear magnetism, you have independent up and down, but the moment you turn on SOC (or non-collinear magnetism) they are no longer independent and are coupled so you have just one DOS. Check for example the eigenvalues in the Ni non-collinear case I added. There is just one bunch of eigenvalues (for single electron states I assume).

ondracka: Please note that for the QE parser is not depending on `n_spin_channels` for DOS parsing, but it has separate codepath to determine the shape...

ndaelman-hu: Yeah, I noticed that the main parser does not extract the DOS

ndaelman-hu: Do you also have the DOS of Ni non-collinear?

ondracka:
> I see. In the future, each code will have its own project.

Haha, so the cycle closes ;-)

electronicparsers/quantumespresso/parser.py
ondracka: I can get some, or I can use some different testcase where non-collinear actually makes sense (like some heavy element where are actually some non-negligible spin orbit effects). But there is only one component in the DOS.

So in general when `n` is the number of electrons
non-spinpolarized calculation: `n/2` two electron bands
collinear spin-polarized:  `n/2` up one electron bands and  `n/2` down one electron bands
noncollinear: `n` one electron bands (distinguishing up and down makes no longer sense)

However, as I said I wanted to add the `n_spin_channels` and added the noncollinear case as a sideeffect just to be sure that I don't parse something incorrectly. I'm not even sure if `n_spin_channels` 4 is a good value to signal noncollinear magnetism and whether we really want to use this field to signal noncollinear magnetism. It could be that we go for 1 (so that DOS, eigenvalues, etc. which detemined the shape according to `n_spin_channels` are happy) and add another extra metainfo field to signal it. In theory we are using four component spinors, but yeah, it doesn't really have the same meaning as in the nonspin-polarized or collinear spin case. What I can do is to leave the field empty and add a test to make sure this is empty until you decide in general how the metainfo for noncollinear calculations should look like. My main worry is just that when data mine for  non-sponpolarized calculations I can do it only with `n_spin_channels = 1` and I want to be sure I don't end with some non-collinear cases as well.

JosePizarro3: For now, this patch could work, but:

`n_spin_channels` shouldn't be more than 2. We should extend the metainfo for adding non-collinear cases in a different way (possibly our way of handling quantum numbers should be able to handle these situations). I wonder even if `n_spin_channels` will survive at all or its purpose will change.

ondracka: I think I'll just add the testcase (with DOS), leave the `n_spin_channels` empty and add a TODO.

ondracka: Add Pt testcase instead of Ni, with DOS. I'm no more parsing the `n_spin_channels` for noncollinear and the tests are checking that its indeed empty. There are also some quick checks for the noncollinear case regarding the dimensionality of eigenvalues and DOS.

tests/test_quantumespressoparser.py
ondracka: Fixes

electronicparsers/quantumespresso/parser.py
ondracka: Fixed the regex

ndaelman-hu (approved):
Okay, thx for applying the requested changes.
We'll see to look into non-collinear calculations in the future: I couldn't get much from the QE docs, yet.

electronicparsers/quantumespresso/parser.py
JosePizarro3: Ok, sounds good to me. I leave @ndaelman-hu to continue with his review (my point was mainly on the physics behind `n_spin_channels`)

ndaelman-hu:
> Okay, thx for applying the requested changes. We'll see to look into non-collinear calculations in the future: I couldn't get much from the QE docs, yet.

@ondracka Feel free to merge (if all threads are resolved).

PR #204, opened 2024-02-26
remove pyscf dep

blueraft:
Removes the pyscf dependency since this doesn't seem to be used anywhere and it causes some issues for MacOS users.


ladinesa (approved):
I thought I already removed this. There was a pr before regarding this. but anyway thanks.

PR #206, opened 2024-02-28
Extend VASP parsing of charges and magnetism

### LINKED ISSUE #174: Magnetic moments in VASP
Add support for atomically-resolved magnetic moments, as well as the total magnetic moment (either directly extracted or derived via normalization).
We should also capture the spin modelling, i.e. collinear or not, magnetically constrained (direction and/or size).
For the time being, we test this out on VASP.

ndaelman-hu:
Fix a bug about capturing the valence electrons ('ZVAL') + add extraction of atomic charges and magentizations.

Closes #174 and Uday's request

### DIFF
--- a/electronicparsers/vasp/parser.py
+++ b/electronicparsers/vasp/parser.py
@@ -66,6 +66,7 @@ from runschema.method import (
 )
 from runschema.calculation import (
     Calculation,
+    Charges,
     Energy,
     EnergyEntry,
     Forces,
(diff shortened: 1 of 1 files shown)


ondracka:
Any chance you can also add parsing of `n_spin_channels` while here (i.e., ISPIN)?

ndaelman-hu:
> Any chance you can also add parsing of `n_spin_channels` while here (i.e., ISPIN)?

Okay, I'll look into it.

ndaelman-hu:
Okay, so the total atomic charges are in. I can also add the projected charges, as well as atomic magnetism, but that'll require extending the schema. @JosePizarro3 We should keep this in mind for when we design `data.ModelOutput`.

Lastly, VASP only reports atomic charges / magnetizations in the `OUTCAR`, but not `vasprun.xml` to my knowledge. Not sure about the new HDF5 format.

PR #214, opened 2024-04-18
Update hdf5 references

ladinesa:


JosePizarro3 (approved):
Very good, thanks for taking care of this.

Btw, this will eventually change with the `HDFDataset` and the new schema. How does it work if we have more than 1 quantity with this type? Does it define a global hdf5 file where it stores this as a dataset? Or it creates as many hdf5 files as needed? I think the second option might not be optimal.

ladinesa:
> Very good, thanks for taking care of this.
> 
> Btw, this will eventually change with the `HDFDataset` and the new schema. How does it work if we have more than 1 quantity with this type? Does it define a global hdf5 file where it stores this as a dataset? Or it creates as many hdf5 files as needed? I think the second option might not be optimal.

There is only one hdf5 for every entry juat like the atchive.

PR #215, opened 2024-04-22
Fix quantity types

ladinesa:
Fix wrong type and shape in metainfo quantities. @TLCFEM

### DIFF
--- a/electronicparsers/abacus/metainfo/abacus.py
+++ b/electronicparsers/abacus/metainfo/abacus.py
@@ -61,7 +61,7 @@ class x_abacus_section_parallel(MSection):
     m_def = Section(validate=False)
 
     x_abacus_nproc = Quantity(
-        type=np.dtype(np.int32),
+        type=np.int32,
         shape=[],
         description="""
         Number of processors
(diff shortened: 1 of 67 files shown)


JFRudzinski (approved):
This seems all fine to me. You may also want @ndaelman-hu to take a quick look in terms of some of the more specific things that you changed.

Question: Is there any case that one would use `np.dtype()` now? I vaguely remember in the past having to use this for some more complicated array structure, but I can't really remember or might be mistaken, or maybe something has changed.

ladinesa:
> This seems all fine to me. You may also want @ndaelman-hu to take a quick look in terms of some of the more specific things that you changed.
> 
> Question: Is there any case that one would use `np.dtype()` now? I vaguely remember in the past having to use this for some more complicated array structure, but I can't really remember or might be mistaken, or maybe something has changed.

You can stil use np.dtype() and it would be equivalent to not using it, it is a matter of preference I guess but @TLCFEM may know more.

ndaelman-hu:
@ladinesa Are there any quantities that you didn't feel sure about? Anything that you want me to check in particular?

ladinesa:
> @ladinesa Are there any quantities that you didn't feel sure about? Anything that you want me to check in particular?

Nothing in particular. I did not make any changes in the schema, simply made sure that the assigned values are compatible with the defs.

PR #220, opened 2024-05-23
np deprecates truth value of np arrays

blueraft:
See [discussion](https://github.com/numpy/numpy/issues/9583)

### DIFF
--- a/electronicparsers/atk/parser.py
+++ b/electronicparsers/atk/parser.py
@@ -326,8 +326,12 @@ class ATKParser:
             sec_atoms.lattice_vectors = atoms.get_cell().array * ureg.angstrom
             sec_atoms.periodic = atoms.get_pbc()
             velocities = atoms.get_velocities()
-            if velocities:
-                sec_atoms.velocities = velocities * (ureg.angstrom / ureg.fs)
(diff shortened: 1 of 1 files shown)


electronicparsers/atk/parser.py
```diff
@@ -326,8 +326,12 @@ def parse_system(atoms):
             sec_atoms.lattice_vectors = atoms.get_cell().array * ureg.angstrom
             sec_atoms.periodic = atoms.get_pbc()
             velocities = atoms.get_velocities()
-            if velocities:
-                sec_atoms.velocities = velocities * (ureg.angstrom / ureg.fs)
+            if isinstance(velocities, np.ndarray):
```
ladinesa: can't we have if velocities is not None here?

blueraft: What do you mean? `if isinstance(velocities, np.ndarray) and velocities is not None`?

blueraft: I see what you mean.

```python
            if velocities is not None:
                sec_atoms.velocities = velocities * (ureg.angstrom / ureg.fs)
```

ladinesa (approved):

PR #223, opened 2024-05-27
Fix gaussian basis set

ladinesa:
A parsing error for gaussian has been reported for this [upload](https://nomad-lab.eu/prod/v1/gui/user/uploads/upload/id/XenO4Le6QyOHXHBpdxtVgA) . The basis set cannot be resolved so I simply set it to None, but @ndaelman-hu you should know more.

The method normalizer also fails so I put up a fix in nomad.

### DIFF
--- a/electronicparsers/gaussian/parser.py
+++ b/electronicparsers/gaussian/parser.py
@@ -1401,6 +1401,7 @@ class GaussianParser:
             self.logger.error(
                 'Found multiple or no basis set', data=dict(n_parsed=len(basis_sets))
             )
+        bs = None
         for basis_set in basis_sets:
             bs = BasisSet(
                 type='gaussians',
@@ -1414,7 +1415,7 @@ class GaussianParser:
             BasisSetContainer(
                 type='atom-centered orbitals',
                 scope=['wavefunction'],
-                basis_set=[bs],
+                basis_set=[bs] if bs is not None else [],
             )
         ]
 
--- a/electronicparsers/gaussian/parser.py
+++ b/electronicparsers/gaussian/parser.py
@@ -2,6 +2,7 @@ import re
 import numpy as np
 import logging
 import ase
+from typing import Optional
 
 from .metainfo import m_env
 
@@ -1306,7 +1307,7 @@ class GaussianParser:
                     name = res[2]
             return prefix, name
 
-        def resolve_basis_set(parameter: str) -> tuple[str, str]:
+        def resolve_basis_set(parameter: str) -> Optional[tuple[str, str]]:
             """This function has 2 responsibilities:
             1. discern `parameter` from other input settings.
             2. verify that `parameter` is a valid basis set name."""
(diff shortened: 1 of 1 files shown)


JosePizarro3:
Hi there,

I cannot see this upload, but is this the last one by Akseli Mansikkamäki (https://nomad-lab.eu/prod/v1/gui/upload/id/MASTxxsySg-SkFffP9j2Zw)? Do you know details on the data? I think he is studying magnetism in charged molecules, but I'd like to make sure.

ndaelman-hu:
> Hi there,
> 
> I cannot see this upload, but is this the last one by Akseli Mansikkamäki (https://nomad-lab.eu/prod/v1/gui/upload/id/MASTxxsySg-SkFffP9j2Zw)? Do you know details on the data? I think he is studying magnetism in charged molecules, but I'd like to make sure.

Neither I can see the upload. @ladinesa Could you ask permissions to share the raw files or have us included?

electronicparsers/gaussian/parser.py
```diff
@@ -1414,7 +1415,7 @@ def resolve_xc_functional(parameter):
             BasisSetContainer(
                 type='atom-centered orbitals',
                 scope=['wavefunction'],
-                basis_set=[bs],
+                basis_set=[bs] if bs is not None else [],
             )
         ]
```
ndaelman-hu: You can leave out the entire `BasisSetContainer` if the requirement stands.

ndaelman-hu (commented):
This is okay for a quick fix. If I could take a look at the error or the raw files, I could probabaly hunt down the real reason.
In the past, there were issues matching the basis set specification due to Gaussian's formatting.

ndaelman-hu:
@ladinesa Okay, I got access to the failing upload. You made the right call: there simply isn't any basis set mentioned.
Likely, this is due to:

 1. this calculation starting from a previous SCF (not uploaded).
 2. this is a frequency calculation, with a different emphasis.
 
 
Still, this is pretty bad practice on Gaussian's side, as SCF routines shouldn't tacitly assume the basis set.
Fro the long-term, I'll see if I can hunt down their default, though Gaussian is likely to have some complex logic in assigning it.

ndaelman-hu:
@JosePizarro3 Regarding the upload you linked, all the Gaussian files parsed well there.
The issue seems to be with Orca, as we don't support CASSCF, yet. I have this on the planned list, but quite low.

We either add a schema there, or we harden the parsing not to fail for unrecognized methods.

ndaelman-hu:
Digging further, the default is [STO-3G](https://gaussian.com/basissets/). I can see to improve on the current correction then.

ladinesa:
@ndaelman there is a problematic test. is it possible for you to look at it so we can merge it already? thanks

ndaelman-hu:
So, some final comments here that will lead to new issues:
1. we need to add missing basis set specs to our tests
2. we will have to extend "GEN" (or "Gen") parsing, as this indicates a user-defined basis set

PR #231, opened 2024-06-20
OpenMX: parse stress tensor

ondracka:
This is quite tricky, because for unknown reasons the stress tensor is not printed into the standard mainfile *.out but only into the standard output. This means, we have to try all files in the directory and check first few lines to see if anything resembles the OpenMX standard output, than do a few consistency check to be extra sure this is a correct stdout corresponding to the mainfile. The parsing itself is trivial. 

Patch written by @jfrodlova with minor tweaks from me.

### DIFF
--- a/electronicparsers/openmx/parser.py
+++ b/electronicparsers/openmx/parser.py
@@ -24,7 +24,7 @@ import numpy as np
 from datetime import datetime
 import re
 import io
-from os import path
+from os import path, listdir
 
 from nomad.datamodel import EntryArchive
 from nomad.units import ureg as units
@@ -37,6 +37,8 @@ from runschema.calculation import (
     BandEnergies,
     Forces,
     ForcesEntry,
+    Stress,
+    StressEntry,
 )
 from runschema.system import AtomsGroup, System, Atoms
 from runschema.method import (
@@ -69,6 +71,20 @@ This is parser for OpenMX DFT code.
 """
 
 # A = (1 * units.angstrom).to_base_units().magnitude
+
+
+def find_stdout(mainfile):
+    mainfile_dir = path.dirname(mainfile)
+    for file in listdir(mainfile_dir):
+        stdout_path = path.join(mainfile_dir, file)
+        if path.isfile(stdout_path):
+            with open(stdout_path) as input_file:
+                for index, line in enumerate(input_file):
+                    if r'Welcome to OpenMX' in line:
+                        return stdout_path
+                    if index == 500:
+                        break
+    return None
 
 
 def _j_mapping() -> dict[tuple[int, int], tuple[float, float]]:
@@ -213,6 +229,7 @@ mainfile_parser = TextParser(
         Quantity(
             'scf.SpinPolarization', r'scf.SpinPolarization\s+(\S+)', repeats=False
         ),
+        Quantity('scf.stress.tensor', r'scf.stress.tensor\s+(.*)', repeats=False),
         Quantity(
             'Atoms.SpeciesAndCoordinates.Unit',
             r'(?i)Atoms.SpeciesAndCoordinates.Unit\s+([a-z]{2,4})',
@@ -252,6 +269,29 @@ mainfile_parser = TextParser(
             r'Eigenvalues \(Hartree\) \S+ SCF KS-eq\.\s+\*{59}\s*\*{59}([^*]+)',
             sub_parser=eigenvalues_parser,
             repeats=False,
+        ),
+        Quantity(
+            'elapsed.time',
+            r'Total Computational Time.*[0-9.]+\s+[0-9.]+\s+[0-9.]+\s+([0-9.]+)',
+        ),
+    ]
+)
+
+stdout_parser = TextParser(
+    quantities=[
+        Quantity(
+            'scf.stress.tensor',
+            r'Stress tensor \(Hartree/bohr\^3\)\s+\*{55}([^*]+)',
+            repeats=True,
+        ),
+        Quantity(
+            'Uele',
+            r'\*{19} MD=\s*\d+\s+SCF=\s*1\s+\*{19}(?:.|\n)*?Uele\s+=\s+(-?\d+\.\d+)',
+            repeats=True,
+        ),
+        Quantity(
+            'elapsed.time',
+            r'Total Computational Time.*[0-9.]+\s+[0-9.]+\s+[0-9.]+\s+([0-9.]+)',
         ),
     ]
 )
@@ -713,6 +753,8 @@ class OpenmxParser:
         # Use the previously defined parsers on the given mainfile
         mainfile_parser.mainfile = mainfile
         mainfile_parser.parse()
+        elapsed_time_mainfile = mainfile_parser.get('elapsed.time')
+
         del mainfile_parser._file_handler
 
         # Get system from the MD file
@@ -767,6 +809,35 @@ class OpenmxParser:
             ignore_md_file = True
             self.logger.warning('.md file does not contain enough MD steps')
 
+        # Stress tensor values are not written to outfile but only to stdout.
+        # That means we have to check if the stdout is actually present and
+        # look for the stress values there.
+        stress_tensor_enabled = mainfile_parser.get('scf.stress.tensor')
+        stdout_uele_list = []
+        stress_tensors = None
+        if stress_tensor_enabled == 'on':
+            # find_stdout returns path to correct stdout if it exists
+            # or None otherwise
+            stdout_path = find_stdout(mainfile)
+
+            if mainfile_md_steps is not None and stdout_path is not None:
+                mainfile_uele = mainfile_md_steps[0].get('SCF')[0].get('Uele')
+                stdout_parser.mainfile = stdout_path
+                stdout_parser.parse()
+                stdout_uele_list = stdout_parser.get('Uele')
+                stdout_uele = stdout_uele_list[0]
+                elapsed_time_stdout = stdout_parser.get('elapsed.time')
+
+                # It is possible there are multiple calculation in the same directory
+                # and thus multiple stdout files as well. We need to find the correct
+                # stdout to look for the stress values. A simple test is to check the
+                # total runtime and later also the uele consistency.
+                if (
+                    elapsed_time_mainfile == elapsed_time_stdout
+                    and abs(stdout_uele - mainfile_uele) < 1e-8
+                ):
+                    stress_tensors = stdout_parser.get('scf.stress.tensor')
+
         if mainfile_md_steps is not None:
             for i, md_step in enumerate(mainfile_md_steps):
                 system = System()
@@ -787,10 +858,19 @@ class OpenmxParser:
                         scf = ScfIteration()
                         sec_calc.scf_iteration.append(scf)
                         u_ele = scf_step.get('Uele')
+
                         if u_ele is not None:
                             scf.energy = Energy(
                                 sum_eigenvalues=EnergyEntry(value=u_ele * units.hartree)
                             )
+
+                if stress_tensors is not None:
+                    stress_matrix = np.array(stress_tensors[i]).reshape(3, 3)
+                    stress_matrix = stress_matrix * units.hartree / units.bohr**3
+                    sec_calc.stress = Stress()
+                    sec_calc.stress.total = StressEntry()
+                    sec_calc.stress.total.value = stress_matrix
+
                 u_tot = md_step.get('Utot')
                 if u_tot is not None:
                     sec_calc.energy = Energy(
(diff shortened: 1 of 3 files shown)


ndaelman-hu:
> This is quite tricky, because for unknown reasons the stress tensor is not printed into the standard mainfile *.out but only into the standard output. This means, we have to try all files in the directory and check first few lines to see if anything resembles the OpenMX standard output [...]

Couldn't we relegate this to the mainfile matching? Where in the file do you have to check?

electronicparsers/openmx/parser.py
```diff
@@ -71,6 +73,20 @@
...
+def find_stdout(mainfile):
+    mainfile_dir = path.dirname(mainfile)
+    for file in listdir(mainfile_dir):
+        stdout_path = path.join(mainfile_dir, file)
+        if path.isfile(stdout_path):
+            with open(stdout_path) as input_file:
+                for index, line in enumerate(input_file):
+                    if r'Welcome to OpenMX' in line:
+                        return stdout_path
+                    if index == 500:
```
ndaelman-hu, line 85: `>`, just to be safe

electronicparsers/openmx/parser.py
```diff
@@ -253,6 +270,29 @@ def convert_eigenvalues(string):
             sub_parser=eigenvalues_parser,
             repeats=False,
         ),
+        Quantity(
+            'elapsed.time',
+            r'Total Computational Time.*[0-9.]+\s+[0-9.]+\s+[0-9.]+\s+([0-9.]+)',
```
ndaelman-hu, line 275: Why leave out the first 3 digits?
Btw, you can write their regex more succinctly as `(?:[0-9.]+\s+){3}`

electronicparsers/openmx/parser.py
```diff
@@ -253,6 +270,29 @@ def convert_eigenvalues(string):
...
+        ),
+        Quantity(
+            'Uele',
+            r'\*{19} MD=\s*\d+\s+SCF=\s*1\s+\*{19}(?:.|\n)*?Uele\s+=\s+(-?\d+\.\d+)',
+            repeats=True,
+        ),
+        Quantity(
+            'elapsed.time',
+            r'Total Computational Time.*[0-9.]+\s+[0-9.]+\s+[0-9.]+\s+([0-9.]+)',
+        ),
```
ndaelman-hu, line 295: I'm seeing multiple keys for elapsed time.
If the code prints this for each subroutine, add some clarification like a comment or modified key. The tree structure should show this too, but it's easy to miss when reviewing or skimming.

electronicparsers/openmx/parser.py
```diff
@@ -767,6 +809,35 @@ def parse(self, mainfile: str, archive: EntryArchive, logger):
             ignore_md_file = True
             self.logger.warning('.md file does not contain enough MD steps')
 
+        # Stress tensor values are not written to outfile but only to stdout.
+        # That means we have to check if the stdout is actually present and
+        # look for the stress values there.
+        stress_tensor_enabled = mainfile_parser.get('scf.stress.tensor')
+        stdout_uele_list = []
+        stress_tensors = None
+        if stress_tensor_enabled == 'on':
```
ndaelman-hu, line 818: Would convert this in the `Quantity` already via `str_operation`.
As such, you can just directly call `mainfile_parser.get('scf.stress.tensor')` here.

electronicparsers/openmx/parser.py
```diff
@@ -767,6 +809,35 @@ def parse(self, mainfile: str, archive: EntryArchive, logger):
...
+            # find_stdout returns path to correct stdout if it exists
+            # or None otherwise
+            stdout_path = find_stdout(mainfile)
+
+            if mainfile_md_steps is not None and stdout_path is not None:
```
ndaelman-hu, line 823: > if mainfile_md_steps is not None

This logic kinda comes out of the blue... Could you add context as a comment?

electronicparsers/openmx/parser.py
```diff
@@ -767,6 +809,35 @@ def parse(self, mainfile: str, archive: EntryArchive, logger):
...
+                mainfile_uele = mainfile_md_steps[0].get('SCF')[0].get('Uele')
+                stdout_parser.mainfile = stdout_path
+                stdout_parser.parse()
+                stdout_uele_list = stdout_parser.get('Uele')
+                stdout_uele = stdout_uele_list[0]
```
ndaelman-hu, line 828: combine into 1 line

tests/test_openmxparser.py
```diff
@@ -315,3 +316,46 @@ def test_graphite(parser):
     assert method.electronic.van_der_waals_method == 'G10'
     assert method.scf.n_max_iteration == 100
     assert method.scf.threshold_energy_change.magnitude == approx(Ha_to_J(1e-8))
+
+    calculation_first = run.calculation[0]
+    stress_tensor_first = calculation_first.stress.total.value
```
ndaelman-hu, line 321: combine into 1 line?

tests/test_openmxparser.py
```diff
@@ -315,3 +316,46 @@ def test_graphite(parser):
...
+                [-0.00000000, 0.00000000, -0.00002298],
+            ]
+        )
+        * units.hartree
+        / units.bohr**3
+    )
+    assert np.allclose(stress_tensor_first, stress_tensor_expected_first, 0.0)
+
+    calculation_last = run.calculation[-1]
+    stress_tensor_last = calculation_last.stress.total.value
```
ndaelman-hu, line 336: same

tests/test_openmxparser.py
```diff
@@ -315,3 +316,46 @@ def test_graphite(parser):
...
+            [
+                [-0.00001778, -0.00000244, -0.00000000],
+                [-0.00000244, -0.00001879, 0.00000000],
+                [-0.00000000, 0.00000000, -0.00002298],
+            ]
+        )
+        * units.hartree
+        / units.bohr**3
+    )
+    assert np.allclose(stress_tensor_first, stress_tensor_expected_first, 0.0)
```
ndaelman-hu, line 333: The (absolute?) tolerance seems quite high... I think that it'll allow too much to pass.
Can you try lowering it a bit?

tests/test_openmxparser.py
```diff
@@ -315,3 +316,46 @@ def test_graphite(parser):
...
+            [
+                [-0.00000003, -0.00000001, 0.00000000],
+                [-0.00000001, -0.00000004, -0.00000000],
+                [0.00000000, -0.00000000, -0.00001568],
+            ]
+        )
+        * units.hartree
+        / units.bohr**3
+    )
+    assert np.allclose(stress_tensor_last, stress_tensor_expected_last, 0.0)
```
ndaelman-hu, line 348: same

ndaelman-hu (approved):
All good, and passes the checks. Pls take a look at the tolerances in the tests: those seem a bit too permissive.

Also pointed out some places where the code can be compacted a bit or a comment with explanation is warranted. Overall, this parser is quite well documented, though.

ondracka:
> > This is quite tricky, because for unknown reasons the stress tensor is not printed into the standard mainfile *.out but only into the standard output. This means, we have to try all files in the directory and check first few lines to see if anything resembles the OpenMX standard output [...]
> 
> Couldn't we relegate this to the mainfile matching? Where in the file do you have to check?

I'm not sure I follow here. You mean we should reuse some parts from the mainfile matching to look for the stdout? Regarding the stdout itself, it looks quite similar to the out file (which we use as a mainfile right now). Some parts are more verbose in the out file, it has a consistent .out suffix, so matching is easy, and as a larger benefit there is also the complete input file reprinted in the so parsing method is much easier with the .out file, which is why I choosed it as a mainfile initially (but stdout for mainfile would work to some extend as well). The downside is, that writing of the out file is heavily buffered, so usually it is written after the calculation finishes (so no outfile for example in case of crash, while the stdout is much better in this case).

So the stress values are effectivelly scattered all over the stdout at the end off corresponding scf cycles.

tests/test_openmxparser.py
ondracka: My bad for not reading the docs that the relative and absolute tolerances are adding up...
I went for  `rtol=1e-8, atol=0.0` in the end.

electronicparsers/openmx/parser.py
ondracka: Yeah, this was halfway baked check to not except later when we index the `mainfile_md_steps`, but we were still missing checks that we actually have scf steps, and more... 
In the end I just wrapped all of this instead in `try: .... except: pass` block instead of the `if mainfile_md_steps is not None and stdout_path is not None:` check.

electronicparsers/openmx/parser.py
ondracka: Actually it looks like this , so we want the last one.

```
                               Min_ID   Min_Time       Max_ID   Max_Time
   Total Computational Time =    11     1110.529          0     1110.545
```

electronicparsers/openmx/parser.py
ondracka: Nice catch, one is for mainfile, one is for outfile, but one was extra ;-)

electronicparsers/openmx/parser.py
ndaelman-hu: I trust you there, it just comes across as weird when reading the parser.
I don't think that you need to match the rest too, just add a comment clarifying.

ndaelman-hu:
> > > This is quite tricky, because for unknown reasons the stress tensor is not printed into the standard mainfile *.out but only into the standard output. This means, we have to try all files in the directory and check first few lines to see if anything resembles the OpenMX standard output [...]
> > 
> > 
> > Couldn't we relegate this to the mainfile matching? Where in the file do you have to check?
> 
> I'm not sure I follow here. You mean we should reuse some parts from the mainfile matching to look for the stdout? Regarding the stdout itself, it looks quite similar to the out file (which we use as a mainfile right now). Some parts are more verbose in the out file, it has a consistent .out suffix, so matching is easy, and as a larger benefit there is also the complete input file reprinted in the so parsing method is much easier with the .out file, which is why I choosed it as a mainfile initially (but stdout for mainfile would work to some extend as well). The downside is, that writing of the out file is heavily buffered, so usually it is written after the calculation finishes (so no outfile for example in case of crash, while the stdout is much better in this case).
> 
> So the stress values are effectivelly scattered all over the stdout at the end off corresponding scf cycles.

You caught on well. Matching the `stdout` file can definitely be done via the system, as we allow for checking the contents via `mainfile_contents_re` (though maybe not 500 lines...). It can exist in parallel with the current matching.
The issue would rather be how to join both in a single archive, as I think atm only 1 would persist.
I'll discuss better approaches for dealing with auxiliary files with @ladinesa when he comes back from holidays, as this system still heavily relies on the parser dev to scan the file.

Your current solutions works though, and I imagine that you need it. That's why this PR got my approval.

ondracka:
Yeah, I think the mainfile matching for the stdout would work if we wanted to do it like for VASP, e.g. to have a preferred mainfile  (vasprun.xml) and than a backup mainfile (but that would be much more work). IIRC there might be some similar issues in VASP (like some properties that are in OUTFILE and not in vasprun.xml or vice versa, not 100% sure how it is handled there, but I think its always just one file used, right). We went this custom way mostly for the reason that this way we only need to jump through hoops if the .out file already indicates the stress values should be there. 

We can rework this when you agree with @ladinesa on the correct solution.

JosePizarro3:
There is a function in utils, `get_files`, which can help you scan for the file you are looking for. This function can scan deeper (i.e., going down or up with respect to the mainfile location) or not in the folder structure for this file. 

Matching files might be a bit tricky when coming to "re-match" it back with the mainfile.

ndaelman-hu:
>  IIRC there might be some similar issues in VASP (like some properties that are in OUTFILE and not in vasprun.xml or vice versa, not 100% sure how it is handled there, but I think its always just one file used, right).

Indeed, VASP supports 3 formats now, but they do not all have exactly the same information content, or even up to the same accuracy...

> There is a function in utils, get_files, which can help you scan for the file you are looking for. This function can scan deeper (i.e., going down or up with respect to the mainfile location) or not in the folder structure for this file.

That's a nice proposal. I guess you're referring to the `utils.py` in `nomad-simulations`, so we should bear this in mind when migrating to the new schema.

Overall, I think this should be a base feature of NOMAD, though. We already have some support for scanning the folder structure at large via the workflow yaml. That's what I'd like to discuss with @ladinesa when he comes back.

@ondracka I wouldn't necessarily wait for this now. Your implementation is fine, so feel free to merge.

JosePizarro3:
> That's a nice proposal. I guess you're referring to the utils.py in nomad-simulations, so we should bear this in mind when migrating to the new schema.

No no, there is a `utils` folder in this repo. We should indeed move this `get_files` function up to the NOMAD source code.

PR #232, opened 2024-06-20
Update frequency parsing in Gaussian

ndaelman-hu:
Limit extraction of vibrational analysis to '--' header. Also update schema and add dynamic unit extraction.

### DIFF
--- a/electronicparsers/gaussian/parser.py
+++ b/electronicparsers/gaussian/parser.py
@@ -391,7 +391,7 @@ class GaussianOutParser(TextParser):
                 unit='debye * angstrom**3',
             ),
             Quantity(
-                'frequencies', r'\n *Frequencies \-\-\s*(.+)', dtype=float, repeats=True
+                'frequencies', r'\n *Frequencies \-+\s*(.+)', dtype=float, repeats=True
             ),
             Quantity(
                 'reduced_masses',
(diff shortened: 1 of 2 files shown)


electronicparsers/gaussian/parser.py
```diff
@@ -100,6 +100,25 @@ def str_to_force_constants(val_in):
             fc = fc + fc.T - np.diag(fc.diagonal())
             return fc
 
+        def str_to_units(unit: str):
```
ladinesa: a map would be more desirable here instead of ifs

ladinesa (approved):

electronicparsers/gaussian/parser.py
ndaelman-hu: I agree, is a bit nicer to look at. Will update.

ndaelman-hu:
@JosePizarro3 my test is failing at Yambo GW and TBStudio... Did something break there, cause I only touched Gaussian?

ndaelman-hu:
Note: won't add testing. Have made a memo in #233 .

ndaelman-hu:
> @JosePizarro3 my test is failing at Yambo GW and TBStudio... Did something break there, cause I only touched Gaussian?

So, these errors come from the updated type checking and are NOT caused by this PR.
@JosePizarro3 and @ladinesa I leave it up to you to fix those at your discretion. This PR will be merged.

PR #235, opened 2024-06-24
229 problematic frequency parsing from gaussian files

ndaelman-hu:
Fix a testing error tracing back to bad unit conversion.


ndaelman-hu:
@ladinesa It's quite annoying that `str_operations` don't allow for error-tracing. Could we see to avoid this during the new parsers?

ladinesa:
sure, I can put some more meaningful logs

PR #257, opened 2024-10-02
Wien2k: core-hole parsing

jfrodlova:


jfrodlova:
cc @ondracka

JFRudzinski:
@jfrodlova sorry to hijack your PR. We recently published this repo on Zenodo. If you would like to be included as an author, please send me your full name and orcid (if you have one).

@ndaelman-hu - I guess you will see this PR, but just in case

jfrodlova:
> @jfrodlova sorry to hijack your PR. We recently published this repo on Zenodo. If you would like to be included as an author, please send me your full name and orcid (if you have one).

Hi @JFRudzinski, thanks for reaching out.
Here is a link to my ORCID page https://orcid.org/0000-0002-3113-9641 (Jitka Frodlová)

ondracka:
@ndaelman-hu Any chance for a review? I've already done some review locally before @jfrodlova submitted it here, but your feedback would be appreciated since you implemented the CoreHole in other parsers.

tests/test_wien2kparser.py
```diff
@@ -121,3 +124,16 @@ def test_dos(parser, caplog):
...
+    parser.parse('tests/data/wien2k/TiN-corehole/TiN-corehole.scf', archive, None)
+
+    sec_method = archive.run[0].method[0]
+    atom_par = sec_method.atom_parameters[0]
+    assert atom_par.atom_index == 2
+    assert atom_par.core_hole.n_electrons_excited == 1
+    assert atom_par.core_hole.l_quantum_number == 1
+    assert atom_par.core_hole.j_quantum_number == 1.5
+    assert atom_par.core_hole.n_quantum_number == 2
+    assert atom_par.core_hole.occupation == 3
```
ndaelman-hu, line 139: I'm noticing now that we should have a better denomination of which orbital / state is being described.
Will see to add that to the new schema.

tests/data/wien2k/TiN-corehole/TiN-corehole.clmcor
ndaelman-hu, line 1: This is a pretty extensive test suite. Do we need all of these files?
If you want to retain a future reference, you can always upload it to NOMAD-lab.

electronicparsers/wien2k/parser.py
```diff
@@ -205,6 +207,42 @@ def get_atoms(self):
...
+                            dtype=float,
+                        ),
+                        Quantity(
+                            'n_quantum_numbers',
+                            r'(\d+), ?-?\d+,[0-9.]+',
+                            repeats=True,
+                            dtype=int,
+                        ),
+                        Quantity(
+                            'kappas',
```
ndaelman-hu, line 235: Does `kappas` refer to the relativistic quantum number, i.e. $\kappa = (l−j)(2j+1)$?

electronicparsers/wien2k/parser.py
```diff
@@ -1053,7 +1096,44 @@ def parse_method(self):
...
+            # http://www.wien2k.at/reg_user/textbooks/usersguide.pdf
+            for atom_index, atom in enumerate(atoms):
+                kappas = atom.get('kappas')
+                n_quantum_numbers = atom.get('n_quantum_numbers')
+                occupancies = atom.get('occupancies')
+                number_of_orbitals = len(kappas)
+                for orbital in range(number_of_orbitals):
+                    max_occupancy = abs(kappas[orbital] * 2)
+                    occupancy = int(occupancies[orbital])
+                    if occupancy < max_occupancy:
```
ndaelman-hu, line 1114: so, `max_occupancy` is not reachable? is this because of the core-hole?
I think you can easily extend the logic to consider a core-hole of 0 electron density being excited.
Sounds silly, but you'd be amazed by what people come up with

electronicparsers/wien2k/parser.py
```diff
@@ -1053,7 +1096,44 @@ def parse_method(self):
...
+                        if kappas[orbital] >= 0:
+                            spin_quantum_number = -1
+                        else:
+                            spin_quantum_number = 1
+                        j_quantum_number = (
+                            -kappas[orbital] / spin_quantum_number - 1 / 2
+                        )
+                        l_quantum_number = j_quantum_number - (spin_quantum_number / 2)
```
ndaelman-hu: Goign off by the $\kappa$ definition in [exciting](http://exciting.wikidot.com/neon-understanding-species), I'm failing to see the math behind this derivation...
Not saying you're wrong, but pls add a comment / link to any kind of docs. Also, add comments for any assumptions made.

electronicparsers/wien2k/parser.py
```diff
@@ -1053,7 +1096,44 @@ def parse_method(self):
...
+                        atom_obj = AtomParameters()
+                        atom_obj.atom_index = atom_index
+                        core_hole = CoreHole()
+                        atom_obj.core_hole = core_hole
+                        j_quantum_number = np.array([j_quantum_number])
+                        core_hole.j_quantum_number = j_quantum_number
+                        core_hole.l_quantum_number = l_quantum_number
+                        core_hole.n_quantum_number = n_quantum_number
+                        core_hole.n_electrons_excited = electrons_excited
+                        core_hole.occupation = occupancy
```
ndaelman-hu: While it's quite clean to seggregate internal variables from thoise being written to the archive, I think that here you can declare the more trivial ones directly when instantiating `CoreHole`.

This is more of a taste thing, so take it for what you want. Alternatively, just add a newline so the responsibility of each code blocks is clearer.

electronicparsers/wien2k/parser.py
```diff
@@ -1053,7 +1096,44 @@ def parse_method(self):
             sec_method.k_mesh = sec_k_mesh
             sec_k_mesh.points = kpoints[0]
             sec_k_mesh.weights = kpoints[1]
-
+        if self.inc_parser.mainfile:
```
ndaelman-hu, line 1100: Is the presence of an `.inc` file sufficient to guarantee a core-hole?
I'd expect some kind of check for a keyword too.

electronicparsers/wien2k/parser.py
ndaelman-hu: Maybe add a comment elaborating.

ndaelman-hu: Btw, do you have any interest in storing `kappa` directly? This was noted as a TODO.

electronicparsers/wien2k/parser.py
ndaelman-hu: Moreover, `CoreHole` comes with a `set_j_quantum_number`. This looks like it would interest you.

tests/test_wien2kparser.py
ndaelman-hu, line 1: Couldn't verify the numerical values.

electronicparsers/wien2k/parser.py
ndaelman-hu, line 1: Pls also consider specifying `CoreHole.dscf_state`. It specifies the kind of approximation made.

ndaelman-hu (commented):
I have a couple of questions that are more so directed to the schema on our side.
Also some minor suggestions, mostly for clarification.

electronicparsers/wien2k/parser.py
ondracka: I don't follow here, if we have 0 electron core-hole, this is just standard groundstate, right? This might be related to a later question, the inc file is always there, the only way to find out that we have a core-hole is by checking for the core-level occupancy. So there is no core-hole keyword of anything.

electronicparsers/wien2k/parser.py
ondracka: Nope, inc file is always present, the only way to find out that we have a core-hole is by checking for the occupancy.

electronicparsers/wien2k/parser.py
ondracka: @jfrodlova , please add this one with a test, the value is always 'final' https://nomad-lab.eu/prod/v1/gui/analyze/metainfo/nomad/section_definitions@nomad.datamodel.metainfo.simulation.method.AtomParameters/core_hole/dscf_state

The tricky part with all-electron code is that the initial state is really just the standard groundstate, there is no way to tell.

electronicparsers/wien2k/parser.py
ondracka: There is already link to the Wien2k user guide (and the following table):

image

From the kappa definition (Wien2k seems to follow the textbook definition)
```
kappa = (l-j)*(2j+1); also j = l + s/2
therefore kappa = (-s/2) * (2*j + 1)
- 2 * kappa / s = 2*j + 1
2 * j = - 2 * kappa / s - 1
j = -kappa/s - 1/2
```

So I believe this should be correct.

electronicparsers/wien2k/parser.py
ondracka: Actually no interest here is storing kappa ;-)

tests/data/wien2k/TiN-corehole/TiN-corehole.clmcor
ondracka: Right, we will purge the large files.

@jfrodlova lets get rid of the TiN-corehole.clm* charge density files as well as lcorepot and lsplcore files.

electronicparsers/wien2k/parser.py
ondracka: @jfrodlova please move it to the instantiating

tests/test_wien2kparser.py
ondracka: I double checked and they are correct (this a Ti2p3/2 single electron core-hole).

ondracka:
Thanks a lot @ndaelman-hu  for the review, much appreciated.

electronicparsers/wien2k/parser.py
ondracka: And the occupancy is unfortunately not written also in the mainfile, so yeah, its kind of stupid to have to go through the input files (with the usual issues like possible inconsistency etc. but I don't see another way).

ndaelman-hu: The main issue with input files is that users might not always upload them. I guess we'll have to signal the relevance here.

electronicparsers/wien2k/parser.py
ndaelman-hu: Thx for writing it out. This is the same as with exciting, but without the absolute value.
I guess I got confused somewhere in the mapping. All good.

electronicparsers/wien2k/parser.py
ndaelman-hu: > This might be related to a later question, the inc file is always there, the only way to find out that we have a core-hole is by checking for the core-level occupancy. So there is no core-hole keyword of anything.

Yes, that was the edge I case I meant. In some codes, you could trigger the core-hole routine, but essentially retain the ground state, but here it seems to be guarded for already.

electronicparsers/wien2k/parser.py
ondracka: Right

@jfrodlova please add a logger warning if the file is not present. Something like
"case.inc file is missing, no corehole information will be parsed (if corehole present)."

electronicparsers/wien2k/parser.py
jfrodlova: Added. Is it what you meant?

electronicparsers/wien2k/parser.py
jfrodlova: Okay, added.

electronicparsers/wien2k/parser.py
jfrodlova: Done.

tests/data/wien2k/TiN-corehole/TiN-corehole.clmcor
jfrodlova: Sure, done.

jfrodlova:
Many thanks for the code review @ndaelman-hu and @ondracka

electronicparsers/wien2k/parser.py
ndaelman-hu: Thx! We've had issues with line-separated logs before. Those were different from yours -namely pystrings-, but could you please double-check that the 2nd phrase is also printed?

ndaelman-hu: Also, but this is just a friendly suggestion: your logic is correct, but it takes up a lot of space between both options. Anyone encountering the `else` statement will have forgotten the matching `if`-statement.

I think that flipping the order of the choices (and inverting the logical checks) will help clarity here.
Of course, clarity is always susceptible to a certain height.

ndaelman-hu (approved):
Thx for applying the feedback!
Just check my comments about the logging, should be 2 secs.

You're good to merge

jfrodlova:
Thanks for your remarks, @ndaelman-hu!
I checked the logger - the whole sentence is printed.

ndaelman-hu:
> Thanks for your remarks, @ndaelman-hu! I checked the logger - the whole sentence is printed.

That's awesome! I learned something new today :) Thx.

PR #260, opened 2024-10-27
added author

JFRudzinski:


JFRudzinski:
@ndaelman-hu do you know why these tests are failing? Can you check the dev branch if not? I didn't change anything relevant here.

ndaelman-hu:
> @ndaelman-hu do you know why these tests are failing? Can you check the dev branch if not? I didn't change anything relevant here.

The failing tests indicate a change no. "calculation" sections in psi4 + a change in how often the ocean method section was accessed. The latter looks like overkill to test, tbh.

Will look into this further this evening.

ndaelman-hu:
> > @ndaelman-hu do you know why these tests are failing? Can you check the dev branch if not? I didn't change anything relevant here.
> 
> The failing tests indicate a change no. "calculation" sections in psi4 + a change in how often the ocean method section was accessed. The latter looks like overkill to test, tbh.
> 
> Will look into this further this evening.

The error in the Psi4 test is a potatoes-potatoes question: it performs 6 calculations, but the last one fails.
Before the failed one was not extracted, now it is. Honestly, I think we can relax the requirements, if we continue this logic in the new schema. There's a case to be made there, but then we need the concept of a "failed calculation".

Still looking into where this change came from.

ndaelman-hu:
@ladinesa Do you know the reason behind measuring `m_mod_count`? Else, I'd remove this test from the _ocean_ parser.

JosePizarro3:
> @ladinesa Do you know the reason behind measuring `m_mod_count`? Else, I'd remove this test from the _ocean_ parser.

Hey, I added these tests. Using `m_mod_count` was my attempt of checking that there are 22 parsed parameters, but if you have a better option, I will be happy to hear it.

But most importantly, why now there are only 21 parsed quantities? Where did the last one go and why?

ndaelman-hu:
> > @ladinesa Do you know the reason behind measuring `m_mod_count`? Else, I'd remove this test from the _ocean_ parser.
> 
> Hey, I added these tests. Using `m_mod_count` was my attempt of checking that there are 22 parsed parameters, but if you have a better option, I will be happy to hear it.
> 
> But most importantly, why now there are only 21 parsed quantities? Where did the last one go and why?

Okay, thank you for clarifying. I guess that you wanted to check whether all `x_ocean_screen` parameters were set.
They are, but I only count 21, not 22, so that's likely the source of the error. Still tracing where this change came from.

Let me verify, but it appears that `m_mod_count` counts the no. of modifications to the section. If the order or setting properties is modified, e.g. a variable is overwritten, this would also change the count. I think telling the dev how to code is too restrictive.
In this case, one could just target the namespace to ensure full coverage, which by itself is a proper test.

JosePizarro3:
> > > @ladinesa Do you know the reason behind measuring `m_mod_count`? Else, I'd remove this test from the _ocean_ parser.
> > 
> > 
> > Hey, I added these tests. Using `m_mod_count` was my attempt of checking that there are 22 parsed parameters, but if you have a better option, I will be happy to hear it.
> > But most importantly, why now there are only 21 parsed quantities? Where did the last one go and why?
> 
> Okay, thank you for clarifying. I guess that you wanted to check whether all `x_ocean_screen` parameters were set. They are, but I only count 21, not 22, so that's likely the source of the error. Still tracing where this change came from.
> 
> Let me verify, but it appears that `m_mod_count` counts the no. of modifications to the section. If the order or setting properties is modified, e.g. a variable is overwritten, this would also change the count. I think telling the dev how to code is too restrictive. In this case, one could just target the namespace to ensure full coverage, which by itself is a proper test.

Yes, I think the best should be using one of the other methods. I don't want to excuse myself, but bear in mind that metainfo.py is one of the worst documented modules, if not the most important one. I guess `m_def.all_quantities` would work better. And I did this back in the day when I was learning NOMAD still.

But in any case, the parser now is not working as before.

ndaelman-hu:
Tracking this down will take too much time, especially given the nature of these tests. The base `metainfo.py` got recently changed, so it might be the source of these changes.

Regardless, I think that fixing the tests as in #216 is more appropriate here.
@JosePizarro3 I don't think that the parsed results changed in the case of `test_oceanparser.py`, just the no. times it got modified. That's why I think that the new formulation will be more robust.

PR #261, opened 2024-10-28
Fix broken tests (Psi4, Ocean)

ndaelman-hu:
Modify the tests as discussed in #260 :

- Psi4: capture failed calculation.
- Ocean: improve metric for full-coverage of `x_ocean_screen_parameters`

### DIFF
--- a/tests/test_oceanparser.py
+++ b/tests/test_oceanparser.py
@@ -18,7 +18,6 @@
 
 import pytest
 import numpy as np
-import re
 
 from nomad.datamodel import EntryArchive
 from electronicparsers.ocean import OceanParser
@@ -65,9 +64,8 @@ def test_tio2(parser):
     assert sec_bse.core_hole.broadening.to('eV').magnitude == approx(0.89)
 
     sec_ocean_screen = sec_method[-1].x_ocean_screen
-    for screen_param in vars(x_ocean_screen_parameters).keys():
-        if re.match('x_ocean_', screen_param):
-            assert getattr(sec_ocean_screen, screen_param) is not None
+    for screen_param in x_ocean_screen_parameters.m_def.quantities:
+        assert sec_ocean_screen.m_is_set(screen_param)
     assert sec_ocean_screen.x_ocean_dft_energy_range == approx(150.0)
 
     # Calculation
--- a/tests/test_oceanparser.py
+++ b/tests/test_oceanparser.py
@@ -18,9 +18,11 @@
 
 import pytest
 import numpy as np
+import re
 
 from nomad.datamodel import EntryArchive
 from electronicparsers.ocean import OceanParser
+from electronicparsers.ocean.metainfo.ocean import x_ocean_screen_parameters
 
 
 def approx(value, abs=0, rel=1e-6):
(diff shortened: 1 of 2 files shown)


ndaelman-hu:
The breaking tests are due to a change in dependency that traces back to line 44 in `metainfo.py` from https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/commit/5d80c14e27e56878d0923b19290b2e8cec9c4d0c

It appears to be specific to Python3.9, since I cannot reproduce it with Python3.10

JFRudzinski (approved):
looks fine to me. You may need to upgrade the typing-extension version, see: https://github.com/nomad-coe/electronic-parsers/pull/260/files

ndaelman-hu:
> looks fine to me. You may need to upgrade the typing-extension version, see: https://github.com/nomad-coe/electronic-parsers/pull/260/files

If you already have this fix, then I don't need to add it: I am merging into your branch.

Will give @JosePizarro3 give a bit more time to also supply a review, if he so wishes.
Else, I merge this evening, around ~18:00h.

tests/test_oceanparser.py
```diff
@@ -61,8 +62,10 @@ def test_tio2(parser):
     assert sec_bse.core_hole.solver == 'Lanczos-Haydock'
     assert sec_bse.core_hole.mode == 'absorption'
     assert sec_bse.core_hole.broadening.to('eV').magnitude == approx(0.89)
+
     sec_ocean_screen = sec_method[-1].x_ocean_screen
-    assert sec_ocean_screen.m_mod_count == 22
```
JosePizarro3, line 65: 🤔 still, not sure why the properties being parsed are now 21 and not 22... is there something wrong with the parser?

tests/test_psi4parser.py
```diff
@@ -135,11 +135,16 @@ def test_ecp_basis(parser):
     assert method[5].electronic.method == 'RHF'
 
     calc = archive.run[0].calculation
-    assert len(calc) == 5
+    assert len(calc) in (5, 6)
```
JosePizarro3, line 138: not sure if I understand this

JosePizarro3 (approved):

tests/test_oceanparser.py
ndaelman-hu: No, there were always 21 properties being parsed (I checked the file history). They are still all being filled.

This section was however originally modified 22 times. This is just speculation on my part, but the 1st modification is likely a clean instantiation. Probably, some kind of optimization is now applied that manages to combine the instantiation with first "set" action? Or maybe instantiation is no longer counted as a "modification"?

This is what I meant with `m_mod_count` being an imperfect stand-in for the filling of the section. The current check directly targets the quantities set.

TLCFEM: Better not to use this internal field.
This is used to track modifications to a particular section, which is further used internally to track necessary recalculation of cached properties.
Previous implementation was not complete, it is revised to have a better track.
Internal things like this is not intended for testing.
Changes of this `m_mod_count` do not necessarily mean something is wrong with the parser.

tests/test_psi4parser.py
ndaelman-hu: There used to be 5 calculations (and 6 systems, 6 methods). Now the 6th calculation is also added. The last step is a failed (and captured) calculation step.*

Since I'm unsure of the exact source of the change (only have a vague idea), I cannot guarantee that all 6 calculations will always be extracted. IF they are, I can enforce that the last one matches the failed step.

I could restrict it to length 6, sure, that would be more robust. That could lead to us having to touch this legacy test various times during the repo's existence.

*: Note that we should add the concept of a failed calculation to the new schema too.

tests/test_oceanparser.py
JosePizarro3: thanks for the info!

ndaelman-hu:
Some old commits from a previous PR got mixed in. I removed them now.
The commits to be merged are those that were reviewed.

PR #263, opened 2024-11-13
Bump depedencies

ndaelman-hu:


ndaelman-hu:
Updated the `[dev]` dependencies to match `nomad-lab`.

ladinesa:
can you perhaps also fix the error associated with the removal of metainfo.Environment

ndaelman-hu:
> can you perhaps also fix the error associated with the removal of metainfo.Environment

But this is on the nomad-lab side. I don't mind removing it, as long as Markus and Theodore agree.
Moreover, is there a certain protocol for a deprecation time period for people to adapt or how to announce removal?

ladinesa:
> > can you perhaps also fix the error associated with the removal of metainfo.Environment
> 
> But this is on the nomad-lab side. I don't mind removing it, as long as Markus and Theodore agree. Moreover, is there a certain protocol for a deprecation time period for people to adapt or how to announce removal?

no i mean if there is still some lingering usage of Environment in the parsers, I remember removing them a few months back. I thought this is the reason for the failure in the tests, but it seems not

ndaelman-hu:
> > > can you perhaps also fix the error associated with the removal of metainfo.Environment
> > 
> > 
> > But this is on the nomad-lab side. I don't mind removing it, as long as Markus and Theodore agree. Moreover, is there a certain protocol for a deprecation time period for people to adapt or how to announce removal?
> 
> no i mean if there is still some lingering usage of Environment in the parsers, I remember removing them a few months back. I thought this is the reason for the failure in the tests, but it seems not

No, the reason for the failure is the magres parser. (Btw, we should check on which schema the standalone plugin depends.)
There seems to be mention of `Environment` in the cp2k parser. Will look into it.

pyproject.toml
```diff
@@ -26,25 +26,25 @@ homepage = "https://github.com/nomad-coe/electronic-parsers"
...
-    'typing-extensions==4.7.1',
-    'ruff==0.1.8'
+    'mypy==1.0.1',
+    'pytest>= 5.3.0, <8',
+    'pytest-timeout>=1.4.2',
+    'pytest-cov>=2.7.1',
+    'ruff>=0.6',
+    'typing-extensions>=4.12',
 ]
 # these are extra deps for nomad infra as a result of using nomad.search
```
blueraft, line 37: You could use `nomad-lab[infrastructure]` here instead.

blueraft (approved):
Thanks!

pyproject.toml
ndaelman-hu: Like this? I kep `mongoengine`, since it's a higher version. I could also remove it, though.

blueraft: Yep looks good

ndaelman-hu: @ladinesa is there any reason for having such a high `mongoengine` version? `0.20` vs `0.4`

ladinesa: maybe adapt the one in nomad, or i think you can simply add [infrastructure] in nomad-lab and you can remove these

ndaelman-hu: it's already in infrastructure, yes

ndaelman-hu:
@ladinesa Should I also open an issue for deprecating

```
# For automatic workflows
                from nomad.search import search
                from nomad.app.v1.models import MetadataRequired
```

That would remove the `nomad-lab[infrastructure]` dependency

pyproject.toml
ladinesa: i am not anymore sure why i did not include all infrastructure packages from nomad. but if it works then better

ladinesa:
> @ladinesa Should I also open an issue for deprecating
> 
> ```
> # For automatic workflows
>                 from nomad.search import search
>                 from nomad.app.v1.models import MetadataRequired
> ```
> 
> That would remove the `nomad-lab[infrastructure]` dependency

i added this in todo, we can adapt my implementation in magres parser where i did not use search

JosePizarro3:
> i added this in todo, we can adapt my implementation in magres parser where i did not use search

please, note Sathya is developing the parser in its own repo and asked to disconnect the parser in here

ladinesa:
> > i added this in todo, we can adapt my implementation in magres parser where i did not use search
> 
> please, note Sathya is developing the parser in its own repo and asked to disconnect the parser in here

we should discuss how to handle external developments, since imo we should have full control over parsers that we interface with nomad

ndaelman-hu:
> > > i added this in todo, we can adapt my implementation in magres parser where i did not use search
> > 
> > 
> > please, note Sathya is developing the parser in its own repo and asked to disconnect the parser in here
> 
> we should discuss how to handle external developments, since imo we should have full control over parsers that we interface with nomad

This is a separate discussion. This PR just updates some dependencies.

Sathya just has to properly set up his `nomad.yaml`, that's all. We are NOT going to start stripping off parsers, just because somebody made their own. I agree that this is an example of a counter-argument against grouping parsers into a single repo, though.

pyproject.toml
ndaelman-hu: Uploaded the Wannier90 test data to my local instance with this branch. It can still generate and find the SP and DMFT workflow.
The magres data produces only a SP, but I think that this due to the data itself.

Will be closing this PR.

PR #270, opened 2024-12-16
Fix cp2k inp

ndaelman-hu:
Fix a loophole in the logic where CP2K tries to access an empty list of (restarted) input files.

Prompted by user Max Graml. Here's a screenshot of the failing logs:
Screenshot from 2024-12-16 18-04-25

### DIFF
--- a/electronicparsers/ams/parser.py
+++ b/electronicparsers/ams/parser.py
@@ -1116,7 +1116,7 @@ class RKFParser(FileParser):
         nspin = general.get('nspin', 1)
         xc_functional = {}
         if (ldapot := general.get('ldapot')) is not None:
-            xc_functional['LDA'] = self._ldapot.get(ldapot)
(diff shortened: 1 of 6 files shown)


ladinesa (approved):

ndaelman-hu:
@ladinesa could we see to roll this out for Max/

ladinesa:
> @ladinesa could we see to roll this out for Max/

i can create a release, but I am sure it will trigger some compatibility issues again in nomad... but I think it is perfect time to have issues since it is the holidays. are you sure you have all fixes in now in develop?

ndaelman-hu:
> but I think it is perfect time to have issues since it is the holidays.

Let's wait for the new year then. I'll inform Max.

> are you sure you have all fixes in now in develop?

The other failed calculations had the same error.

ladinesa:
please advise when I should make a release.

PR #271, opened 2025-01-16
Patch QE `n_electrons`

ndaelman-hu:
A user pointed out that with spin-polarization turned on, the QE parser extracts 3 values for the total electron count and attempts to assign them, causing an error:

```
"exception":string"Traceback (most recent call last): File "/home/cecilia/lavoro/nomad-distro-dev-aa/packages/nomad-FAIR/nomad/processing/data.py", line 1508, in parsing parser.parse( File "/home/cecilia/lavoro/nomad-distro-dev-aa/packages/nomad-FAIR/nomad/parsing/parser.py", line 460, in parse self.mainfile_parser.parse(mainfile, archive, logger) File "/home/cecilia/lavoro/nomad-distro-dev-aa/.venv/lib/python3.12/site-packages/electronicparsers/quantumespresso/parser.py", line 3563, in parse self.parse_method(run) File "/home/cecilia/lavoro/nomad-distro-dev-aa/.venv/lib/python3.12/site-packages/electronicparsers/quantumespresso/parser.py", line 3499, in parse_method sec_method.electronic.n_electrons = number_of_electrons ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/home/cecilia/lavoro/nomad-distro-dev-aa/packages/nomad-FAIR/nomad/metainfo/metainfo.py", line 1090, in __setattr__ return super().__setattr__(name, value) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/home/cecilia/lavoro/nomad-distro-dev-aa/packages/nomad-FAIR/nomad/metainfo/metainfo.py", line 727, in wrapper return method(self, obj, value, **kwargs) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/home/cecilia/lavoro/nomad-distro-dev-aa/packages/nomad-FAIR/nomad/metainfo/metainfo.py", line 3273, in __set__ value = self.type.normalize(value, section=obj, **kwargs) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/home/cecilia/lavoro/nomad-distro-dev-aa/packages/nomad-FAIR/nomad/metainfo/data_type.py", line 546, in normalize return super().normalize(value, **kwargs) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/home/cecilia/lavoro/nomad-distro-dev-aa/packages/nomad-FAIR/nomad/metainfo/data_type.py", line 294, in normalize raise ValueError( ValueError: Cannot set [7. 4. 3.] for the scalar quantity runschema.method.Electronic.n_electrons:Quantity."
"timestamp":string"2025-01-16 15:53.12"
"level":string"ERROR"
}
```

### DIFF
--- a/electronicparsers/quantumespresso/parser.py
+++ b/electronicparsers/quantumespresso/parser.py
@@ -2037,7 +2037,12 @@ class QuantumEspressoOutParser(TextParser):
             ),
             Quantity(
                 'number_of_electrons',
-                rf'number of electrons\s*=\s*({re_float})\s*(?:\(up:\s*({re_float})\s*,\s*down:\s*({re_float}))?',
+                rf'number of electrons\s*=\s*({re_float})',
+                dtype=float,
+            ),
+            Quantity(
+                'number_of_spin_electrons',
+                rf'number of electrons\s*=[\s\S]*up:\s*({re_float})\s*,\s*down:\s*({re_float})',
                 dtype=float,
             ),
             Quantity(
@@ -3489,16 +3494,10 @@ class QuantumEspressoParser:
                 if atom_sp[i] is not None:
                     setattr(sec_method_atom_kind, atom_species_names[i], atom_sp[i])
 
-        n_electrons = run.get_header('number_of_electrons', [])
-        if len(n_electrons) == 3:
-            sec_method.electronic.n_electrons = n_electrons[0]
-        elif len(n_electrons) == 0:
-            pass
-        else:
-            self.logger.warning(
-                'Corrupted extraction `n_electrons`. Cannot set value.',
-                n_electrons=n_electrons,
-            )  # this error could be better managed if we retianed the og keys
+        if (n_electrons := run.get_header('number_of_electrons')) is not None:
+            sec_method.electronic.n_electrons = n_electrons
+        elif len(n_spin_electrons := list(run.get_header('number_of_spin_electrons', []))) == 2:
+            sec_method.electronic.n_electrons = sum(n_spin_electrons)
 
     def init_parser(self):
         self.out_parser.mainfile = self.filepath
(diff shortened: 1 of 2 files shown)


ndaelman-hu:
I now split the extraction between the total count and the one divided by spin-channel.
The latter is used is when the former is missing or corrupted.

electronicparsers/quantumespresso/parser.py
```diff
@@ -2037,7 +2037,12 @@ def str_to_sticks(val_in):
             ),
             Quantity(
                 'number_of_electrons',
-                rf'number of electrons\s*=\s*({re_float})\s*(?:\(up:\s*({re_float})\s*,\s*down:\s*({re_float}))?',
```
ladinesa, line 2040: can you send me the text which fails for this regex pattern? as far as i can see this is generic enough for both cases. we should try to keep it in one quantity.

ndaelman-hu: The pattern isn't failing itself. The schema only needs the 1st match. There was a bug where the parser passed along all 3, violating the shape. The new shape rules actually brought it too light. So we should maybe consider reprocessing QE calcs...

If for some God-foresaken reason the 1st number can't be extracted, it can still be reconstructed from the latter 2.
It's very niche, I can remove that logic.

ladinesa: can you then simply put this logic in swtting the value based on the length of the list

ndaelman-hu: I tried that first (it's in an older commit). The answer's no, since idk which value(s) is missing when the list is of length 2 or 1. The only choice is to support this corruption recovery or not. Up to you.

ladinesa: we can then make this as a sub-parser to resolve the electron type, this i think is still a better way than introducing multiple quantites

ladinesa (approved):

PR #277, opened 2025-02-25
Fix projections, fermi energy, and added new method for getting band structure from wannier90_hr.dat file

16-vikrant:
### Fix
**Problem**: The fermi energy and projection are not correctly parsed in the current implementation. 
**Solution**: I have modified the projection quantity which could only parse projections written as 
```Begin Projections
Pb:p
I:p
End Projections 
```
Now, we can also parse projections written in the form
```
Begin Projections
Pb:p
I:p
c=4.419090763607875,0.3192596584266408,9.629367383527137:l=0;l=1   !  1  C
c=8.083085628608565,5.309565669515625,2.5067956004325:l=0;l=1   !  2  N
End Projections
```
The Fermi energy was set to the onsite hopping energy at `0 0 0` site. I have modified the parser to get it from `wannier90.win` if the value is preset. If `wannier90.win` does not have Fermi energy, half of onsite hopping energy at `0 0 0` site will be used which is correct for many semiconductors. I have logged a user warning in this case. 

### New feature
In current implementation, band structure is generated using `wannier90_band.dat` file. I have implemented a new module which can calculate band structure from hopping parameters written in `wannier90_hr.dat` file.


ndaelman-hu:
@16-vikrant could you also supply an example raw output file? It would help with checking the regex.

16-vikrant:
@ndaelman-hu Find attached a `wannier90.win` file. 
[wannier90.zip](https://github.com/user-attachments/files/19007103/wannier90.zip)

electronicparsers/wannier90/parser.py
```diff
@@ -144,15 +145,32 @@ def init_quantities(self):
...
+                if '!' in proj:
+                    element_match = re.search(r'[!#]\s+\d+\s+([A-Za-z]+)', proj)
+                    l_values_match = re.findall(r'l=(\d+)', proj)
+                    if element_match and l_values_match:
+                        element = element_match.group(1)
+                        orbitals = [l_to_orbital[l] for l in l_values_match if l in l_to_orbital]
+                        # print(element, orbitals)
+                        cleaned_proj = f"{element}:{', '.join(orbitals)}"
+                        # replace the original line with the cleaned one
+                        val_n[val_n.index(proj)] = cleaned_proj
```
ndaelman-hu: You don't have to funnel `cleaned_proj` through the list comprehension. Feel free to just return a new list following the same format.

On that note: how does the format of this list compare with the original list?

ndaelman-hu: Looking into the handling logic for the momentum below (line 413), note that the processed result is split on `;`, whereas you already split it here. You could leverage the logic below by simply substituting out `l=`.

ndaelman-hu: Finally, while you are only extending the parser's features, I think that a new version should use subparser with more quantities. This is especially apparent from the projections. It would also simplify the handling logic.

electronicparsers/wannier90/parser.py
```diff
@@ -447,17 +465,128 @@ def parse_hoppings(self):
...
-            n_wigner_seitz_points_half = int(
-                0.5 * sec_hopping_matrix.n_wigner_seitz_points
+
+            if self.win_parser.get('energy_fermi') is not None:
+                # Setting Fermi level to the value in the win file
+                energy_fermi = self.win_parser.get('energy_fermi') * ureg.eV # If win is parsed before hopping and energy_fermi is found
+                sec_scc_energy.fermi = energy_fermi
+                sec_scc_energy.highest_occupied = energy_fermi
+            else:
+                # Setting Fermi level to the first orbital onsite energy (This may be inaccurate)
```
ndaelman-hu: What kinds of cases lead you to the disclaimer?

electronicparsers/wannier90/parser.py
```diff
@@ -447,17 +465,128 @@ def parse_hoppings(self):
...
             sec_scc.energy = sec_scc_energy
-            # Setting Fermi level to the first orbital onsite energy
-            n_wigner_seitz_points_half = int(
-                0.5 * sec_hopping_matrix.n_wigner_seitz_points
+
+            if self.win_parser.get('energy_fermi') is not None:
+                # Setting Fermi level to the value in the win file
+                energy_fermi = self.win_parser.get('energy_fermi') * ureg.eV # If win is parsed before hopping and energy_fermi is found
+                sec_scc_energy.fermi = energy_fermi
+                sec_scc_energy.highest_occupied = energy_fermi
```
ndaelman-hu: this common line may stand outside the `if-else` block

electronicparsers/wannier90/parser.py
```diff
@@ -447,17 +465,128 @@ def parse_hoppings(self):
         try:
             sec_scc_energy = Energy()
             sec_scc.energy = sec_scc_energy
-            # Setting Fermi level to the first orbital onsite energy
-            n_wigner_seitz_points_half = int(
-                0.5 * sec_hopping_matrix.n_wigner_seitz_points
+
+            if self.win_parser.get('energy_fermi') is not None:
+                # Setting Fermi level to the value in the win file
+                energy_fermi = self.win_parser.get('energy_fermi') * ureg.eV # If win is parsed before hopping and energy_fermi is found
```
ndaelman-hu: if you set the default parameter in `get`, you don't need the `if-else` statement anymore.
just define `n_wigner_seitz_points_half` in a local method

electronicparsers/wannier90/parser.py
```diff
@@ -447,17 +465,128 @@ def parse_hoppings(self):
...
+                )
+                energy_fermi = (
+                    0.5 * sec_hopping_matrix.value[n_wigner_seitz_points_half][0][5] * ureg.eV
+                )
+                sec_scc_energy.fermi = energy_fermi
+                sec_scc_energy.highest_occupied = energy_fermi
+                self.logger.warning(
+                    'User should verify the Fermi level. It is set to the half of first orbital onsite energy. This may be inaccurate.'
+                )
+        except Exception:
```
ndaelman-hu: pylint should've also pointed this out: capturing generic `Exception`s is bad practice. try to drill them down to the cases that you can handle

electronicparsers/wannier90/parser.py
```diff
@@ -447,17 +465,128 @@ def parse_hoppings(self):
...
+            return
+        return energy_fermi
+    
+    def parse_tb_bands(self):
```
ndaelman-hu, line 505: pydocs

electronicparsers/wannier90/parser.py
```diff
@@ -447,17 +465,128 @@ def parse_hoppings(self):
...
+        if not hr_files:
+            return
+        
+        hoppings = self.hr_parser.get('hoppings')
+        if isinstance(hoppings, (list, np.ndarray)) and len(hoppings) >= 7:
+            real_e = hoppings[5::7] 
+            imag_e = hoppings[6::7] 
+        num_wann = self.hr_parser.get('degeneracy_factors')[0]
+        n_w_s_p = self.hr_parser.get('degeneracy_factors')[1]
+        degen_factors = self.hr_parser.get('degeneracy_factors')[2:]
```
ndaelman-hu:
```python
num_wann, n_w_s_p, *degen_factors = self.hr_parser.get('degeneracy_factors')
```

electronicparsers/wannier90/parser.py
```diff
@@ -447,17 +465,128 @@ def parse_hoppings(self):
...
+        (n_kpoints, band_segments_points, kpoints) = self.get_k_points()
```
ndaelman-hu: parentheses unnecessary

electronicparsers/wannier90/parser.py
```diff
@@ -447,17 +465,128 @@ def parse_hoppings(self):
...
+        kpoints_ = np.vstack(kpoints)
+
+        hmnr = np.zeros((n_w_s_p, num_wann, num_wann), dtype=complex)
```
ndaelman-hu: better variable name than `hmnr`?

electronicparsers/wannier90/parser.py
```diff
@@ -447,17 +465,128 @@ def parse_hoppings(self):
...
+        hoppings_reshaped = hoppings[:rows * 7].reshape((rows, 7))
+        real_vec = hoppings_reshaped[::num_wann*num_wann, :3]
+
+        for i in range(n_w_s_p):
+            for j in range(num_wann):
+                for k in range(num_wann):
+                    index = i * num_wann * num_wann + j * num_wann + k
+                    hmnr[i, j, k] = (
+                        real_e[index] + 1j * imag_e[index]
+                    ) / degen_factors[i]
```
ndaelman-hu: Please use `numpy` here. Suggested rewrite (double-check):
```python
# Reshape the 1D arrays into 3D arrays
real_e_3d = real_e.reshape(n_w_s_p, num_wann, num_wann)
imag_e_3d = imag_e.reshape(n_w_s_p, num_wann, num_wann)

# Create complex array and divide by degen_factors using broadcasting
hmnr = (real_e_3d + 1j * imag_e_3d) / degen_factors[:, np.newaxis, np.newaxis]
```

electronicparsers/wannier90/parser.py
```diff
@@ -447,17 +465,128 @@ def parse_hoppings(self):
...
+            htb = np.zeros((num_wann, num_wann), dtype=complex)
+            for i in range(n_w_s_p):
+                phase_factor = np.exp(2.0 * np.pi * 1j * np.dot(k, real_vec[i]))
+                htb += hmnr[i, :, :] * phase_factor
+            return htb    
+
+        # Compute the band structure
+        bands = []
+        for i in kpoints_:
+            bands.append(np.sort(eigvals(hamiltonian_tb(i)).real))
```
ndaelman-hu: why are we diagonalizing Hamiltonians..?

ndaelman-hu: list comprehension + variable name `kpoint_`

ndaelman-hu: Or use numpy:

```python
def hamiltonian_tb_vectorized(k):
    phase_factors = np.exp(2.0 * np.pi * 1j * np.dot(real_vec, k))
    return np.sum(hmnr * phase_factors[:, np.newaxis, np.newaxis], axis=0)

def compute_bands(k):
    return np.sort(eigvals(hamiltonian_tb_vectorized(k)).real)

compute_bands_vec = np.vectorize(compute_bands, signature='(n)->(m)')
bands = compute_bands_vec(kpoints_)
```

electronicparsers/wannier90/parser.py
```diff
@@ -447,17 +465,128 @@ def parse_hoppings(self):
...
+            sec_k_band_segment.n_kpoints = band_segments_points[n]
+            sec_k_band_segment.kpoints = kpoints[n]
+
+            bkp_last = bkp_init + band_segments_points[n]
+            energies = np.reshape(
+                bands[bkp_init:bkp_last, :], (n_spin, band_segments_points[n], n_bands)
+            )
+            #self.logger.warning(energies)
+            # print(energies)
+            occs = np.reshape(
```
ndaelman-hu:
```python
occs = np.where(energies < energy_fermi_eV, 2.0, 0.0)
```

electronicparsers/wannier90/parser.py
ndaelman-hu, line 1: Please also remove the commented-out code. I suppose it was originally used for debugging.

ndaelman-hu (approved):
Thx for extending the range of features. Only readability remarks, no actual bugs spotted, so  somewhere between "Approved" and "Comment".

I've marked some places where `numpy` can give a clearer (and faster) result than vanilla Python. Also consider streamlining the projection logic.

16-vikrant:
Hello @ndaelman-hu 

I have improved the code based on your comments and suggestions. Could you please review it again?

Thank you!

electronicparsers/wannier90/parser.py
```diff
@@ -144,16 +145,38 @@ def init_quantities(self):
             ),
         ]
 
-
+import re
```
ndaelman-hu: leave some newlines between imports and the class definition

electronicparsers/wannier90/parser.py
16-vikrant: I have modified this part of the code. Now, the code returns a new list with same format. 

https://github.com/16-vikrant/electronic-parsers/commit/cefb02e1f87171079bb024f430731297749ba538#diff-6dc3e5f21e59a27bfb3140365ac43425dbf50796983b4835b5f28b31ad5c222aR174

electronicparsers/wannier90/parser.py
16-vikrant: The logic is inherently flawed. I have removed this part. Now, the parser will get the `fermi energy` from `wannier90.win` file and set it to zero if not found. 

https://github.com/16-vikrant/electronic-parsers/commit/cefb02e1f87171079bb024f430731297749ba538#diff-6dc3e5f21e59a27bfb3140365ac43425dbf50796983b4835b5f28b31ad5c222aR473

electronicparsers/wannier90/parser.py
16-vikrant: The loop is removed and we are using `try` and `except`. The parser now `try` to get fermi energy from wannier90.win and set it to zero in `except` block.

electronicparsers/wannier90/parser.py
16-vikrant: `n_wigner_seitz_points_half` is not more used in the code. It was used in fermi energy calculation.

electronicparsers/wannier90/parser.py
16-vikrant: I have removed this part.

electronicparsers/wannier90/parser.py
16-vikrant: I have not written pydocs yet. I will do it soon.

electronicparsers/wannier90/parser.py
16-vikrant: Thanks for the suggestion. I have implemented and tested this in current commit.

electronicparsers/wannier90/parser.py
16-vikrant: Thanks. I have removed the parantheses.

electronicparsers/wannier90/parser.py
16-vikrant: Now the variable name is changed to `hamiltonian_mn_real`.  I agree `hmnr` was not a very clever choice.

electronicparsers/wannier90/parser.py
16-vikrant: Thank you for the suggestion. I have rewritten this part. 

https://github.com/16-vikrant/electronic-parsers/commit/cefb02e1f87171079bb024f430731297749ba538#diff-6dc3e5f21e59a27bfb3140365ac43425dbf50796983b4835b5f28b31ad5c222aR515

electronicparsers/wannier90/parser.py
16-vikrant: This is a great question. We are not always diagonizing Hamiltonians. The parser will get the band structure from `wannier90_band.dat` file. If `wannier90_band.dat` is missing, parser will calculate bandstructure from `hopping matrix` elements given in `wannier90_hr.dat` file. We are working with the wannier90_hr.dat file as part of high-throughput Wannier function calculations.

16-vikrant: Thank you for the suggestion. I have implemented and tested this. 

https://github.com/16-vikrant/electronic-parsers/commit/cefb02e1f87171079bb024f430731297749ba538#diff-6dc3e5f21e59a27bfb3140365ac43425dbf50796983b4835b5f28b31ad5c222aR518

electronicparsers/wannier90/parser.py
16-vikrant: Thank you! This part is also updated now. 

https://github.com/16-vikrant/electronic-parsers/commit/cefb02e1f87171079bb024f430731297749ba538#diff-6dc3e5f21e59a27bfb3140365ac43425dbf50796983b4835b5f28b31ad5c222aR567

electronicparsers/wannier90/parser.py
ndaelman-hu: But `Exception` is still here..?

electronicparsers/wannier90/parser.py
ndaelman-hu: ~~It's strange: I'm still seeing the old version here. Only see your new one in the link. Pls verify that this commit is included in the PR.~~

electronicparsers/wannier90/parser.py
```diff
@@ -447,17 +470,105 @@ def parse_hoppings(self):
...
-                sec_hopping_matrix.value[n_wigner_seitz_points_half][0][5] * ureg.eV
-            )
+            energy_fermi = self.win_parser.get('energy_fermi') * ureg.eV # If win is parsed before hopping and energy_fermi is found
             sec_scc_energy.fermi = energy_fermi
             sec_scc_energy.highest_occupied = energy_fermi
         except Exception:
+            self.logger.warning(
+                'Could not find the Fermi energy in the wannier90.win file. Setting it to zero.'
+            )
+            energy_fermi = 0.0 * ureg.eV
```
ndaelman-hu: just `energy_fermi = 0.0` suffices

electronicparsers/wannier90/parser.py
```diff
@@ -626,11 +754,18 @@ def parse_scc(self):
...
+
         # Wannier90 hoppings section
         self.parse_hoppings()
 
         # Wannier band structure
-        self.parse_bandstructure()
+        if wband_files:
+            self.parse_bandstructure()
+        elif hr_files:
+            self.parse_tb_bands()
```
ndaelman-hu, line 768: Are these 2 mutually exclusive? What if somebody adds them in the same upload directory?

electronicparsers/wannier90/parser.py
```diff
@@ -398,21 +413,63 @@ def fract_cart_sites(sec_atoms, units, val):
...
+                                angular_momentum.append(fallback_label)
+                        except Exception:
+                            # If parsing mr fails (likely missing mr), fallback to pure l
+                            try:
+                                lmom = int(orb.replace('l=', ''))
+                                fallback_map = {0: 's', 1: 'p', 2: 'd', 3: 'f'}
+                                fallback_label = fallback_map.get(lmom, f'l={lmom}')
+                                angular_momentum.append(fallback_label)
+                            except Exception:
+                                self.logger.warning(f'Could not parse orbital label: {orb}')
```
ndaelman-hu, line 440: This may also activate due to failing `l`. If this logic is relatively independent from `mr`, why not perform them in separate code blocks?

electronicparsers/wannier90/parser.py
```diff
@@ -398,21 +413,63 @@ def fract_cart_sites(sec_atoms, units, val):
...
+                    else:
+                        # Direct orbital label (e.g. 'px', 's')
                         angular_momentum.append(orb)
+
                 sec_atom_parameters.orbitals = np.array(angular_momentum)
             except Exception:
                 self.logger.warning('Projected orbital labels not found from win.')
 
+    def get_or_set_fermi_energy(self, default=0.0 * ureg.eV):
```
ndaelman-hu, line 449: Add a comment with a justification for this default

electronicparsers/wannier90/parser.py
```diff
@@ -398,21 +413,63 @@ def fract_cart_sites(sec_atoms, units, val):
...
+            if energy_fermi_value is None:
+                raise ValueError('Fermi energy not found in wannier90.win file')
+            energy_fermi = energy_fermi_value * ureg.eV
+        except Exception as e:
+            self.logger.warning(f'{e}. Setting Fermi energy to default: {default.magnitude} eV.')
+            energy_fermi = default
+
+        sec_scc.energy.fermi = energy_fermi
+        sec_scc.energy.highest_occupied = energy_fermi
+        return energy_fermi
```
ndaelman-hu, line 470: I would EITHER set the relevant fields OR return the value (which is then set in the calling code block).
I've never seen a _get-or-set_ method, you typically pick either.

Case and point: your first try-except block does not exhibit the same setting behavior...

electronicparsers/wannier90/parser.py
```diff
@@ -444,20 +501,92 @@ def parse_hoppings(self):
                 self.logger.warning(
                     'Could not parse the hopping matrix values. Please, revise your output files.'
                 )
+    
+    def parse_tb_bands(self):
+        hr_files = get_files('*hr.dat', self.filepath, self.mainfile)
+        if not hr_files:
+            return
+        if len(hr_files) > 1:
+            self.logger.warning('Multiple hopping data files found.')
```
ndaelman-hu, line 510: Also state what the next step / decision is.

```suggestion
            self.logger.warning('Multiple hopping data files found. Picking the first file found.')
```

electronicparsers/wannier90/parser.py
```diff
@@ -444,20 +501,92 @@ def parse_hoppings(self):
...
+            sec_k_band.reciprocal_cell = (
+                self.archive.run[-1].system[0].atoms.lattice_vectors_reciprocal
             )
-            sec_scc_energy.fermi = energy_fermi
-            sec_scc_energy.highest_occupied = energy_fermi
         except Exception:
-            return
+            self.logger.warning(
+                'Reciprocal cell in band_structure_electronic not set up.'
+            )
```
ndaelman-hu, line 567: Also specify here what your next step is. You just skip the assignment?
Is there no way to recover? Maybe open up an issue where we check afterwards.

ndaelman-hu (approved):
I mostly checked the error-handling, for the parsing logic I defer to you.
A test would be nice, if possible.

If some of our discussion (old or new) takes too much time, just mark it in a follow-up issue.
You can merge if you have working tests (locally or in this repo).

PR #279, opened 2025-03-04
276 add bandstructure support for qe

### LINKED ISSUE #276: Add bandstructure support for QE
The user submitting this issue provided the following workflow layout as an example for producing a band structure plot:

```bash
pw.x < 1-si.scf.in > 1-si.scf.out
pw.x < 2-si.band.in > 2-si.band.out
bands.x < 4-si.bands.in > 4-si.bands.out
plotband.x < 5-si.plotband.in
```

The output of `bands.x`is two-fold: `4-si.bands.out` and `nibands.dat`. The former is more verbose (clear header) and formatted. It also lists the symmetry group of each point, making identification of high-symmetry points (or changing symmetry) easier.

ndaelman-hu:
close #276

### DIFF
--- a/electronicparsers/__init__.py
+++ b/electronicparsers/__init__.py
@@ -41,7 +41,7 @@ class EntryPoint(ParserEntryPoint):
     metadata: Optional[dict] = Field(
         None,
         description="""
-        Metadata passed to the UI. Deprecated. """
+        Metadata passed to the UI. Deprecated. """,
     )
 
     def load(self):
--- a/electronicparsers/abacus/parser.py
+++ b/electronicparsers/abacus/parser.py
@@ -315,7 +315,7 @@ class ABACUSOutParser(TextParser):
                 data = []
                 for i in range(nks):
                     kx, ky, kz, npws = re.search(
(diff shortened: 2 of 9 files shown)


ladinesa:
it appears that Band is a another module, in order to be consistent with the other qe nodule parsers, i suggest we make this also a separate parser. we unify them in the new parsers.

ndaelman-hu:
> it appears that Band is a another module, in order to be consistent with the other qe nodule parsers, i suggest we make this also a separate parser. we unify them in the new parsers.

Yes, but the bandstructure calculation requires both a PWSCF calculation, as well as, a BAND calculation.
I'm fine splitting off BAND into its own module, but where should the processing of their combination occur?
Moreover, do you have an example of another module?

ladinesa:
> > it appears that Band is a another module, in order to be consistent with the other qe nodule parsers, i suggest we make this also a separate parser. we unify them in the new parsers.
> 
> Yes, but the bandstructure calculation requires both a PWSCF calculation, as well as, a BAND calculation. I'm fine splitting off BAND into its own module, but where should the processing of their combination occur? Moreover, do you have an example of another module?

the other parsers are xspectra epw and phonon. they also require a pwscf calulation

ndaelman-hu:
Migrated to https://github.com/nomad-coe/workflow-parsers/pull/55

PR #282, opened 2025-03-19
input validation for atomic species from VASP xml output

kubanmar:
This addresses the issue described here: https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/issues/1850  
The VASP code apparently set the name of the chemical species incorrectly, such that in some cases the descriptions of the elements in the structure section are wrong, e.g. here: https://nomad-lab.eu/prod/v1/gui/search/calculations/entry/id/zlkl1vpS3QHliZfRRxU856Q8sTRF .

Previously, these cases were not detected. I added input validation, by checking that all elements in the section system are actual atom names. If that is not true, a log message is written and it tries to recover by reading the element name from the pseudopotential description. If that also fails another log message is written.

@ndaelman-hu , you looked into this before: Would this fix also solve the other cases you have seen?

### DIFF
--- a/electronicparsers/vasp/parser.py
+++ b/electronicparsers/vasp/parser.py
@@ -2000,6 +2000,8 @@ class VASPParser:
                         atom_index = self.parser.atom_info.get('atomtypes', {}).get('element', []).index(_label)
                         # get the label from the PP info. The second entry appears to be the chemical element.
                         recovered_label = self.parser.atom_info.get('atomtypes', {}).get('pseudopotential')[atom_index][1]
+                        # account for cases where PP names include underscores
+                        recovered_label = recovered_label.split('_')[0]
                         assert recovered_label in self.allowed_species, 'Recovered species label is also invalid'
                         sec_atoms.labels[idx] = recovered_label
                     except Exception as e:
--- a/electronicparsers/vasp/parser.py
+++ b/electronicparsers/vasp/parser.py
@@ -1642,6 +1642,7 @@ class VASPParser:
             'GW0R': 'scGW0',
             'GWR': 'scGW',
         }
+        self.allowed_species = set(ase.data.chemical_symbols[1:])
 
     def init_parser(self, filepath, logger):
         self.parser = (
@@ -1990,19 +1991,19 @@ class VASPParser:
             sec_atoms.periodic = [True] * 3
             sec_atoms.labels = self.parser.atom_info.get('atoms', {}).get('element', [])
             # create a set of allowed species for faster lookup. The ase data starts with a vacancy labelled 'X', so start with the second entry.
-            allowed_species = set(ase.data.chemical_symbols[1:])
-            unidentified_labels = [(idx, label) for idx, label in enumerate(sec_atoms.labels) if label not in allowed_species]
+            unidentified_labels = [(idx, label) for idx, label in enumerate(sec_atoms.labels) if label not in self.allowed_species]
             if len(unidentified_labels) > 0:
-                self.logger.warning(f'Unidentified atom labels: {unidentified_labels}')
+                self.logger.warning('Unidentified atom labels.', data=list(zip(*unidentified_labels))[1])
                 for idx, _label in unidentified_labels:
                     try:
                         # get the atom index of wrong species
                         atom_index = self.parser.atom_info.get('atomtypes', {}).get('element', []).index(_label)
-                        # get the label from the PP info
-                        # the second entry appears to be the chemical element; there may be a more sustainable way to get this info
-                        sec_atoms.labels[idx] = self.parser.atom_info.get('atomtypes', {}).get('pseudopotential')[atom_index][1]
-                    except Exception:
-                        self.logger.error(f'Unable to recover atom label {_label}.')
+                        # get the label from the PP info. The second entry appears to be the chemical element.
+                        recovered_label = self.parser.atom_info.get('atomtypes', {}).get('pseudopotential')[atom_index][1]
+                        assert recovered_label in self.allowed_species, 'Recovered species label is also invalid'
+                        sec_atoms.labels[idx] = recovered_label
+                    except Exception as e:
+                        self.logger.error('Unable to recover atom label.', data=_label, reason=str(e))
 
             positions = structure.get('positions', None)
             if positions is not None:
--- a/electronicparsers/vasp/parser.py
+++ b/electronicparsers/vasp/parser.py
@@ -1989,6 +1989,20 @@ class VASPParser:
 
             sec_atoms.periodic = [True] * 3
             sec_atoms.labels = self.parser.atom_info.get('atoms', {}).get('element', [])
+            # create a set of allowed species for faster lookup. The ase data starts with a vacancy labelled 'X', so start with the second entry.
+            allowed_species = set(ase.data.chemical_symbols[1:])
+            unidentified_labels = [(idx, label) for idx, label in enumerate(sec_atoms.labels) if label not in allowed_species]
+            if len(unidentified_labels) > 0:
+                self.logger.warning(f'Unidentified atom labels: {unidentified_labels}')
+                for idx, _label in unidentified_labels:
+                    try:
+                        # get the atom index of wrong species
+                        atom_index = self.parser.atom_info.get('atomtypes', {}).get('element', []).index(_label)
+                        # get the label from the PP info
+                        # the second entry appears to be the chemical element; there may be a more sustainable way to get this info
+                        sec_atoms.labels[idx] = self.parser.atom_info.get('atomtypes', {}).get('pseudopotential')[atom_index][1]
+                    except Exception:
+                        self.logger.error(f'Unable to recover atom label {_label}.')
 
             positions = structure.get('positions', None)
             if positions is not None:


electronicparsers/vasp/parser.py
```diff
@@ -1989,6 +1989,20 @@ def parse_system(n_calc):
 
             sec_atoms.periodic = [True] * 3
             sec_atoms.labels = self.parser.atom_info.get('atoms', {}).get('element', [])
+            # create a set of allowed species for faster lookup. The ase data starts with a vacancy labelled 'X', so start with the second entry.
+            allowed_species = set(ase.data.chemical_symbols[1:])
```
ladinesa: maybe add this in init

electronicparsers/vasp/parser.py
```diff
@@ -1989,6 +1989,20 @@ def parse_system(n_calc):
...
+            unidentified_labels = [(idx, label) for idx, label in enumerate(sec_atoms.labels) if label not in allowed_species]
+            if len(unidentified_labels) > 0:
+                self.logger.warning(f'Unidentified atom labels: {unidentified_labels}')
```
ladinesa: remove the f-string format and include `unindentiied_labels` to warning.data

electronicparsers/vasp/parser.py
```diff
@@ -1989,6 +1989,20 @@ def parse_system(n_calc):
...
+                for idx, _label in unidentified_labels:
+                    try:
+                        # get the atom index of wrong species
+                        atom_index = self.parser.atom_info.get('atomtypes', {}).get('element', []).index(_label)
+                        # get the label from the PP info
+                        # the second entry appears to be the chemical element; there may be a more sustainable way to get this info
+                        sec_atoms.labels[idx] = self.parser.atom_info.get('atomtypes', {}).get('pseudopotential')[atom_index][1]
+                    except Exception:
+                        self.logger.error(f'Unable to recover atom label {_label}.')
```
ladinesa: same here

electronicparsers/vasp/parser.py
```diff
@@ -1989,6 +1989,20 @@ def parse_system(n_calc):
...
+            allowed_species = set(ase.data.chemical_symbols[1:])
+            unidentified_labels = [(idx, label) for idx, label in enumerate(sec_atoms.labels) if label not in allowed_species]
+            if len(unidentified_labels) > 0:
+                self.logger.warning(f'Unidentified atom labels: {unidentified_labels}')
+                for idx, _label in unidentified_labels:
+                    try:
+                        # get the atom index of wrong species
+                        atom_index = self.parser.atom_info.get('atomtypes', {}).get('element', []).index(_label)
+                        # get the label from the PP info
+                        # the second entry appears to be the chemical element; there may be a more sustainable way to get this info
```
ladinesa: is this a TODO?

electronicparsers/vasp/parser.py
```diff
@@ -1989,6 +1989,20 @@ def parse_system(n_calc):
 
             sec_atoms.periodic = [True] * 3
             sec_atoms.labels = self.parser.atom_info.get('atoms', {}).get('element', [])
+            # create a set of allowed species for faster lookup. The ase data starts with a vacancy labelled 'X', so start with the second entry.
+            allowed_species = set(ase.data.chemical_symbols[1:])
+            unidentified_labels = [(idx, label) for idx, label in enumerate(sec_atoms.labels) if label not in allowed_species]
+            if len(unidentified_labels) > 0:
+                self.logger.warning(f'Unidentified atom labels: {unidentified_labels}')
+                for idx, _label in unidentified_labels:
```
ndaelman-hu, line 1997: indentation under `if`-statement not really necessary

electronicparsers/vasp/parser.py
```diff
@@ -1989,6 +1989,20 @@ def parse_system(n_calc):
...
+                    try:
+                        # get the atom index of wrong species
+                        atom_index = self.parser.atom_info.get('atomtypes', {}).get('element', []).index(_label)
+                        # get the label from the PP info
```
ndaelman-hu: Just be careful: there is no formal placeholder for elements in PP. There are known spelling mistakes or appendices.
I don't remember anymore whether they are filtered out.

electronicparsers/vasp/parser.py
```diff
@@ -1989,6 +1989,20 @@ def parse_system(n_calc):
 
             sec_atoms.periodic = [True] * 3
             sec_atoms.labels = self.parser.atom_info.get('atoms', {}).get('element', [])
+            # create a set of allowed species for faster lookup. The ase data starts with a vacancy labelled 'X', so start with the second entry.
```
ndaelman-hu, line 1993: I think that @JFRudzinski extended the range of permissible elements, e.g. ghost atoms.

electronicparsers/vasp/parser.py
kubanmar: I did in the new commit. Did it go to the right place?

electronicparsers/vasp/parser.py
kubanmar: Done

electronicparsers/vasp/parser.py
kubanmar: Also done. I added also a reason for the failure.

electronicparsers/vasp/parser.py
kubanmar: I removed the TODO.

electronicparsers/vasp/parser.py
kubanmar: Why not? I can skip the whole block without initializing the iterator of an empty list if the code is indented.

electronicparsers/vasp/parser.py
kubanmar: I added another check of the recovered label. It now falls back to the original wrong label, which should be easier to identify

electronicparsers/vasp/parser.py
kubanmar: I think ghost atoms are not treated here. Is this in conflict with ghost atoms @JFRudzinski ?

ladinesa (approved):

electronicparsers/vasp/parser.py
JFRudzinski: Sorry I missed the first tag.

I don't really understand the context here...You can currently use the `X` label if you have some particle that is not a chemical element. `X` is the ase notation for unknown element. 

I added some fixes to the atomutils so that some of the functions, e.g., getting the summed mass, do not break with `X`. And I also implemented some checks in the topology normalizer for to avoid issues.

kubanmar: Here we just parse the atom type from the VASP xml output. If the fix (not allowing for an `X` in the VASP mainfile listing of elements) will not lead to flagging cases where this may be intended as errors, we can resolve the thread and merge.

JFRudzinski: Oh yes, sorry @kubanmar I forgot about our previous discussion on this. From what I know, I agree that `X` labels found in VASP parsing can be treated as errors...we can confirm with Nathan briefly at our meeting, then you can merge

electronicparsers/vasp/parser.py
ndaelman-hu: can this be resolved then?

electronicparsers/vasp/parser.py
ndaelman-hu: Can you add a test for the underscore case like shown here: https://www.vasp.at/wiki/index.php/Available_pseudopotentials ?

kubanmar: I added a filtering step for the species name

ndaelman-hu: okay, but I would consider adding a test too

electronicparsers/vasp/parser.py
```diff
@@ -1989,6 +1990,22 @@ def parse_system(n_calc):
...
+                    try:
+                        # get the atom index of wrong species
+                        atom_index = self.parser.atom_info.get('atomtypes', {}).get('element', []).index(_label)
+                        # get the label from the PP info. The second entry appears to be the chemical element.
+                        recovered_label = self.parser.atom_info.get('atomtypes', {}).get('pseudopotential')[atom_index][1]
+                        # account for cases where PP names include underscores
+                        recovered_label = recovered_label.split('_')[0]
+                        assert recovered_label in self.allowed_species, 'Recovered species label is also invalid'
+                        sec_atoms.labels[idx] = recovered_label
+                    except Exception as e:
```
ndaelman-hu, line 2007: Generally, it's good practice to specify `Exception`. The only failure modes I see here are `IndexError` and `AssertionError`.

ndaelman-hu (approved):
1 minor comment

electronicparsers/vasp/parser.py
kubanmar: This is an edgecase of an edgecase and we still get warnings or errors if something goes wrong. I don't think we need an automatic test for it.

electronicparsers/vasp/parser.py
kubanmar: Here, by explicitly stating the general `Exception`, I want to indicate that the logging should be done in any case, and no uncaught exceptions should be raised.

electronicparsers/vasp/parser.py
ndaelman-hu: Having `Li_GW` and the like are most definitely not edge cases, but standard naming conventions...

This is outside the scope of this PR, but I do think that the new parsers should better track the provenance of the symbols:

- if it comes from the `POSCAR`, it's the user's naming convention (https://www.vasp.at/wiki/index.php/POSCAR).
- if it comes from the PP, it's defined by its creator.

While the POTCAR order determines the actual species, the user's naming convention often follows it.

kubanmar: Sorry, I was not very clear there. The code in this PR is only executed if the species is not correct in the `element` attribute. I have tested it locally (including the underscore notation) and I don't think writing an explicit testcase for this is necessary. If the species name recovery fails, we will get respective warning log entries.

We maybe should discuss the parsing in general at a different occasion, probably there is more room for improvement.

PR #286, opened 2025-04-03
Parsing Yambo cell parameters 

emolteni:
Parsing Yambo cell parameters from ns.db file, fixing issue about matching n. atoms  n. coordinate lines

Reporting here our changes:

   def parse_input(self):
        if self.mainfile_parser.cpu_files_io.input is None:
            return

        run = self.archive.run[-1]

        # read structure data from netcdf input file
        self.netcdf_parser.mainfile = os.path.join(
            self.maindir, self.mainfile_parser.cpu_files_io.input.file
        )
        if self.netcdf_parser.mainfile is not None:
            system = System()
            run.system.append(system)
            positions = self.netcdf_parser.get('ATOM_POS', [])
            max_n_atoms = self.netcdf_parser.get('MAX_ATOMS')
            n_atoms = self.netcdf_parser.N_ATOMS
            atom_numbers = np.hstack(
                [
                    [self.netcdf_parser.atomic_numbers[int(n)]] * int(n_atoms[int(n)])
                    for n in range(len(n_atoms))
                ]
            )

#     we define the process_and_select function within the parse_input function
            def process_and_select(positions, max_n_atoms, n_atoms): 

                positions = np.array(positions)
                blocks = []
                selected = []

                positions = positions.reshape(-1, 3)
    
                n_points = positions.shape[0] 
                n_blocks = int( int(n_points) // int(max_n_atoms) )

                for i in range(n_blocks):
                    start_idx = int(i) * int(max_n_atoms)
                    end_idx = (int(i) + 1) * int(max_n_atoms)
                    block = positions[start_idx:end_idx]
                    blocks.append(block)
    
    
                for i, block in enumerate(blocks):
                    n_to_select = int(n_atoms[int(i)])
                    selected_from_block = block[:n_to_select]
                    for point in selected_from_block:
                        selected.append(point)
    
                positions=np.array(selected)
                
                return positions        

            positions = process_and_select(positions, max_n_atoms, n_atoms)
            
            system.atoms = Atoms(
                positions = positions * ureg.bohr,
                labels=[chemical_symbols[int(n)] for n in atom_numbers],
            )


ladinesa:
Please try to remove the changes in the quantum espresso parser and and also the backup parser which was i think unintentionally added.

ladinesa:
I also already merged the fix for the spin-polarized non local band energies recently, not sure why they are still in this pr.

electronicparsers/quantumespresso/parser.py
```diff
@@ -16,13 +16,15 @@
...
 #
+import glob
 import logging
 import numpy as np
 import re
 from datetime import datetime
 import os
 from typing import Optional
 
+from nomad.datamodel.metainfo.simulation.calculation import BandStructure
```
ladinesa: you should should the one from runschema.calculation

electronicparsers/quantumespresso/parser.py
```diff
@@ -1631,8 +1633,7 @@
         ],
         'xc_section_method': {
             'x_qe_xc_inlc_name': 'vdwx',
-            'x_qe_xc_inlc_comment': 'vdW-DF-x (reserved Thonhauser,'
-            ' not implemented)',
+            'x_qe_xc_inlc_comment': 'vdW-DF-x (reserved Thonhauser, not implemented)',
```
ladinesa, line 1639: not sure about if this was a max column length issue

electronicparsers/quantumespresso/parser.py
ladinesa: but this is not part o the pr

electronicparsers/quantumespresso/parser.py_BACK
ladinesa, line 1: remoe this file

electronicparsers/quantumespresso/parser.py
ladinesa, line 1: remove this change

electronicparsers/yambo/metainfo/yambo.py
```diff
@@ -341,6 +357,27 @@ class x_yambo_bare_xc_bandenergies(runschema.calculation.BandEnergies):
         """,
     )
 
+####
```
ladinesa: remove comment

electronicparsers/yambo/metainfo/yambo.py
```diff
@@ -341,6 +357,27 @@ class x_yambo_bare_xc_bandenergies(runschema.calculation.BandEnergies):
...
+    ),
+    
+    x_yambo_simulation_cell = Quantity(
+            type=np.float64,
+            shape = [],
+            description="""
+            simulation cell
+        """,
+    ),
+###
```
ladinesa: also this

electronicparsers/yambo/parser.py
```diff
@@ -259,7 +263,14 @@ def init_quantities(self):
                                         rf'\<\d+\|nlXC\|\d+\> *= *({re_f}) *{re_f} \<\d+\|lXC\|\d+\> *= *({re_f}) *{re_f}',
                                         repeats=True,
                                         dtype=np.dtype(np.float64),
-                                    )
+                                    ),
+                                    Quantity(
```
ladinesa, line 263: not sure why this is still in this pr

electronicparsers/yambo/parser.py
```diff
@@ -710,23 +731,59 @@ def parse_input(self):
             system = System()
             run.system.append(system)
             positions = self.netcdf_parser.get('ATOM_POS', [])
+            max_n_atoms = self.netcdf_parser.get('MAX_ATOMS')
```
ladinesa: default for this? probaby put a return statement if this is None?

electronicparsers/yambo/parser.py
```diff
@@ -710,23 +731,59 @@ def parse_input(self):
...
             n_atoms = self.netcdf_parser.N_ATOMS
             atom_numbers = np.hstack(
                 [
                     [self.netcdf_parser.atomic_numbers[int(n)]] * int(n_atoms[int(n)])
                     for n in range(len(n_atoms))
                 ]
             )
+
+#     we define the process_and_select function within the parse_input function
```
ladinesa: describe what this does in as a function docstring

electronicparsers/yambo/parser.py
```diff
@@ -710,23 +731,59 @@ def parse_input(self):
...
+            def process_and_select(positions, max_n_atoms, n_atoms): 
+
+                positions = np.array(positions)
+                blocks = []
+                selected = []
+
+
```
ladinesa: remove line

electronicparsers/yambo/parser.py
```diff
@@ -710,23 +731,59 @@ def parse_input(self):
...
+                n_points = positions.shape[0] 
+                n_blocks = int( int(n_points) // int(max_n_atoms) )
+
+                for i in range(n_blocks):
+                    start_idx = int(i) * int(max_n_atoms)
+                    end_idx = (int(i) + 1) * int(max_n_atoms)
+                    block = positions[start_idx:end_idx]
+                    blocks.append(block)
+    
+
```
ladinesa: remove line

electronicparsers/yambo/parser.py
```diff
@@ -710,23 +731,59 @@ def parse_input(self):
...
+                    n_to_select = int(n_atoms[int(i)])
+                    selected_from_block = block[:n_to_select]
+                    for point in selected_from_block:
+                        selected.append(point)
+    
+                positions=np.array(selected)
+                
+                return positions        
+
+            positions = process_and_select(positions, max_n_atoms, n_atoms)
```
ladinesa: there are possible several failure points in the function, so put a wrap this in a try statement

electronicparsers/yambo/parser.py
```diff
@@ -791,11 +848,16 @@ def parse_local_xc_nonlocal_fock(self, module):
             if len(calc.x_yambo_local_xc_nonlocal_fock_bandenergies) == 0:
                 band_energy = x_yambo_local_xc_nonlocal_fock_bandenergies()
                 calc.x_yambo_local_xc_nonlocal_fock_bandenergies.append(band_energy)
-                sx, vxc = np.transpose([c.band for c in source.corrections])
-                band_energy.x_yambo_sx = np.reshape(sx, (1, *np.shape(sx.T))) * ureg.eV
-                band_energy.x_yambo_vxc = (
-                    np.reshape(vxc, (1, *np.shape(vxc.T))) * ureg.eV
+                nspin = 2 if source.corrections[0].band_sp else 1
```
ladinesa, line 825: i already included this change in a previous pr

ladinesa:
Thanks for the changes, i only have minor comments, is it possible also for you to add tests for this?

emolteni:
> Please try to remove the changes in the quantum espresso parser and and also the backup parser which was i think unintentionally added.

These in the QE parser were just our first tries in that direction: I know the issue of QE band str plotting is now being addressed by the FAIRmat team.   This holds also for your following comments on spin-polarized GW corrections in the Yambo parser (which you have implemented in the meantime).  The only relevant changes to be considered are those I have reported in my initial message in this PR, within the Yambo parser.  The rest has to be ignored, as it's only due to the fact that I have not done sync fork since a long time.

electronicparsers/yambo/parser.py
emolteni: I think I had copied it "by hand" from your PR on this issue, but anyway it seems to be the same as the corresponding part of code in the current general NOMAD.

electronicparsers/yambo/parser.py
emolteni: There seems to be no specified default there for positions, n_atoms ....  either.   They are always all present in every ns.db Yambo file.

electronicparsers/yambo/parser.py
ladinesa: i suggest you rebase and perhaps this will resolve it.

electronicparsers/yambo/parser.py
ladinesa: default for positions is empty list and i think for n_atoms 0, just defensive so it does not raise any error in case failures occur

electronicparsers/yambo/parser.py
emolteni: What is exactly your suggestion about this part? (lines 747-748, and lines 759-760)?  If we just remove these lines, than blocks will be undefined in  "for i, block in enumerate(blocks)".    We need to split the initial (wrong) coordinate matrix into blocks, one for each chemical species, each block having max_n_atoms lines, and only then we can extract for each block  n_atoms lines (where n_atoms in principle assumes different values for the different chemical species present).

ladinesa: no i mean just rhe blanks. try running ruff check --fix

electronicparsers/yambo/parser.py
emolteni: We have now wrapped the whole function in a very general try statement

electronicparsers/yambo/parser.py
emolteni: This can be ignored, and replaced with the current version in the general NOMAD.

emolteni:
> Thanks for the changes, i only have minor comments, is it possible also for you to add tests for this?

Here are some test folders: CH4 and R_methylox are molecules, each with different n. of atoms of different chemical species;  hBN instead is a case with same number of atoms for the 2 chemical species present:

[CH4_correct_folder_str.zip](https://github.com/user-attachments/files/19605897/CH4_correct_folder_str.zip)
[R_methylox_with_db.zip](https://github.com/user-attachments/files/19605910/R_methylox_with_db.zip)
[hBN_1db.zip](https://github.com/user-attachments/files/19605916/hBN_1db.zip)

ladinesa:
> > Thanks for the changes, i only have minor comments, is it possible also for you to add tests for this?
> 
> Here are some test folders: CH4 and R_methylox are molecules, each with different n. of atoms of different chemical species; hBN instead is a case with same number of atoms for the 2 chemical species present:
> 
> [CH4_correct_folder_str.zip](https://github.com/user-attachments/files/19605897/CH4_correct_folder_str.zip) [R_methylox_with_db.zip](https://github.com/user-attachments/files/19605910/R_methylox_with_db.zip) [hBN_1db.zip](https://github.com/user-attachments/files/19605916/hBN_1db.zip)

can you maybe try to add the the test thenselves in test_yambo_parser.py.

ladinesa:
also please remove the QE paraer changes for this pr. i suggest you create another branch in your repo to include only the yambo changes and make this as the source branch for this pr.

emolteni:
> > > Thanks for the changes, i only have minor comments, is it possible also for you to add tests for this?
> > 
> > 
> > Here are some test folders: CH4 and R_methylox are molecules, each with different n. of atoms of different chemical species; hBN instead is a case with same number of atoms for the 2 chemical species present:
> > [CH4_correct_folder_str.zip](https://github.com/user-attachments/files/19605897/CH4_correct_folder_str.zip) [R_methylox_with_db.zip](https://github.com/user-attachments/files/19605910/R_methylox_with_db.zip) [hBN_1db.zip](https://github.com/user-attachments/files/19605916/hBN_1db.zip)
> 
> can you maybe try to add the the test thenselves in test_yambo_parser.py.


I have just uploaded  /CH4_correct_folder_str (as a folder, including both r* file and a /SAV> with a ns.db file) and R_methylox_with_db as a zip (also with both r* and /SAVE with ns.db; I have tried to upload this also unzipped, but it is larger than 25 Mb).   We have not inserted references to these new test folders in test_yambo_parser.py, since we are not sure about the syntax to be used.

ladinesa:
> > > > Thanks for the changes, i only have minor comments, is it possible also for you to add tests for this?
> > > 
> > > 
> > > Here are some test folders: CH4 and R_methylox are molecules, each with different n. of atoms of different chemical species; hBN instead is a case with same number of atoms for the 2 chemical species present:
> > > [CH4_correct_folder_str.zip](https://github.com/user-attachments/files/19605897/CH4_correct_folder_str.zip) [R_methylox_with_db.zip](https://github.com/user-attachments/files/19605910/R_methylox_with_db.zip) [hBN_1db.zip](https://github.com/user-attachments/files/19605916/hBN_1db.zip)
> > 
> > 
> > can you maybe try to add the the test thenselves in test_yambo_parser.py.
> 
> I have just uploaded /CH4_correct_folder_str (as a folder, including both r* file and a /SAV> with a ns.db file) and R_methylox_with_db as a zip (also with both r* and /SAVE with ns.db; I have tried to upload this also unzipped, but it is larger than 25 Mb). We have not inserted references to these new test folders in test_yambo_parser.py, since we are not sure about the syntax to be used.


Follow one of the tests essentially parse the file and assert the quantity.  I suggest you add only one set of test data and should be as minimal as possible if you can generate one

electronicparsers/yambo/parser.py
emolteni: I have just changed the max_atoms line like this:

max_n_atoms = self.netcdf_parser.get('MAX_ATOMS', 0)

..although we are not really sure of the correct syntax for these statements, e.g. why it is different for position and for n_atom.

emolteni:
> > > > > Thanks for the changes, i only have minor comments, is it possible also for you to add tests for this?
> > > > 
> > > > 
> > > > Here are some test folders: CH4 and R_methylox are molecules, each with different n. of atoms of different chemical species; hBN instead is a case with same number of atoms for the 2 chemical species present:
> > > > [CH4_correct_folder_str.zip](https://github.com/user-attachments/files/19605897/CH4_correct_folder_str.zip) [R_methylox_with_db.zip](https://github.com/user-attachments/files/19605910/R_methylox_with_db.zip) [hBN_1db.zip](https://github.com/user-attachments/files/19605916/hBN_1db.zip)
> > > 
> > > 
> > > can you maybe try to add the the test thenselves in test_yambo_parser.py.
> > 
> > 
> > I have just uploaded /CH4_correct_folder_str (as a folder, including both r* file and a /SAV> with a ns.db file) and R_methylox_with_db as a zip (also with both r* and /SAVE with ns.db; I have tried to upload this also unzipped, but it is larger than 25 Mb). We have not inserted references to these new test folders in test_yambo_parser.py, since we are not sure about the syntax to be used.
> 
> Follow one of the tests essentially parse the file and assert the quantity. I suggest you add only one set of test data and should be as minimal as possible if you can generate one


We are trying to write the part about our test in the test_yamboparser.py.  What is the meaning of e.g. calc[1], calc[2] ..?   For some other quantities, we think it's the "shape", but e.g. by comparing line 158  calc[0].x_yambo_filled_bands[1]   with line 195      assert calc[1].x_yambo_filled_bands[1] == 8,  why does calc have two different shapes in the two cases?   Thanks

ladinesa:
> the



> > > > > > Thanks for the changes, i only have minor comments, is it possible also for you to add tests for this?
> > > > > 
> > > > > 
> > > > > Here are some test folders: CH4 and R_methylox are molecules, each with different n. of atoms of different chemical species; hBN instead is a case with same number of atoms for the 2 chemical species present:
> > > > > [CH4_correct_folder_str.zip](https://github.com/user-attachments/files/19605897/CH4_correct_folder_str.zip) [R_methylox_with_db.zip](https://github.com/user-attachments/files/19605910/R_methylox_with_db.zip) [hBN_1db.zip](https://github.com/user-attachments/files/19605916/hBN_1db.zip)
> > > > 
> > > > 
> > > > can you maybe try to add the the test thenselves in test_yambo_parser.py.
> > > 
> > > 
> > > I have just uploaded /CH4_correct_folder_str (as a folder, including both r* file and a /SAV> with a ns.db file) and R_methylox_with_db as a zip (also with both r* and /SAVE with ns.db; I have tried to upload this also unzipped, but it is larger than 25 Mb). We have not inserted references to these new test folders in test_yambo_parser.py, since we are not sure about the syntax to be used.
> > 
> > 
> > Follow one of the tests essentially parse the file and assert the quantity. I suggest you add only one set of test data and should be as minimal as possible if you can generate one
> 
> We are trying to write the part about our test in the test_yamboparser.py. What is the meaning of e.g. calc[1], calc[2] ..? For some other quantities, we think it's the "shape", but e.g. by comparing line 158 calc[0].x_yambo_filled_bands[1] with line 195 assert calc[1].x_yambo_filled_bands[1] == 8, why does calc have two different shapes in the two cases? Thanks

Hi, thanks for writing the tests. the calculation sections are the results parsed from the different modules in the mainfile. The filled bands are read following the line in the mainfile: `States summary         : Full        Metallic    Empty`. l195 and l158 are different simply because they are from different files, Al and GaSb.

emolteni:
> > the
> 
> > > > > > > Thanks for the changes, i only have minor comments, is it possible also for you to add tests for this?
> > > > > > 
> > > > > > 
> > > > > > Here are some test folders: CH4 and R_methylox are molecules, each with different n. of atoms of different chemical species; hBN instead is a case with same number of atoms for the 2 chemical species present:
> > > > > > [CH4_correct_folder_str.zip](https://github.com/user-attachments/files/19605897/CH4_correct_folder_str.zip) [R_methylox_with_db.zip](https://github.com/user-attachments/files/19605910/R_methylox_with_db.zip) [hBN_1db.zip](https://github.com/user-attachments/files/19605916/hBN_1db.zip)
> > > > > 
> > > > > 
> > > > > can you maybe try to add the the test thenselves in test_yambo_parser.py.
> > > > 
> > > > 
> > > > I have just uploaded /CH4_correct_folder_str (as a folder, including both r* file and a /SAV> with a ns.db file) and R_methylox_with_db as a zip (also with both r* and /SAVE with ns.db; I have tried to upload this also unzipped, but it is larger than 25 Mb). We have not inserted references to these new test folders in test_yambo_parser.py, since we are not sure about the syntax to be used.
> > > 
> > > 
> > > Follow one of the tests essentially parse the file and assert the quantity. I suggest you add only one set of test data and should be as minimal as possible if you can generate one
> > 
> > 
> > We are trying to write the part about our test in the test_yamboparser.py. What is the meaning of e.g. calc[1], calc[2] ..? For some other quantities, we think it's the "shape", but e.g. by comparing line 158 calc[0].x_yambo_filled_bands[1] with line 195 assert calc[1].x_yambo_filled_bands[1] == 8, why does calc have two different shapes in the two cases? Thanks
> 
> Hi, thanks for writing the tests. the calculation sections are the results parsed from the different modules in the mainfile. The filled bands are read following the line in the mainfile: `States summary : Full Metallic Empty`. l195 and l158 are different simply because they are from different files, Al and GaSb.

Ok, thanks.  We have just finished writing the part about our CH4 test in the test_yamboparser.py, although we are not completely sure about the meaning of sizes and shapes of some quantities x_yambo_parameters[0].

emolteni:
Hi, please consider that, since this morning, we have started making new changes in the yambo parser, with the aim of trying to let NOMAD plot _spectra_ from Yambo calculations, and therefore completely independent and unrelated wrt to the subject of our pull request you are examining (which is about reporting cell parameters, chemical composition etc).  Should we make another fork, or what else, in order to continue our new (tentative) development for spectra, while you are examining our PR?   Thanks.

ladinesa:
> Hi, please consider that, since this morning, we have started making new changes in the yambo parser, with the aim of trying to let NOMAD plot _spectra_ from Yambo calculations, and therefore completely independent and unrelated wrt to the subject of our pull request you are examining (which is about reporting cell parameters, chemical composition etc). Should we make another fork, or what else, in order to continue our new (tentative) development for spectra, while you are examining our PR? Thanks.

Hi, I suggest you finalize this pr and then i merge it. Then you can rebase your branch and work on the new chenges. Again, please remove the QE parser stuff from this pr, either by creating a new branch and updating the source branch of this pr or simply not commiting your changes to the QE parser. Thanks!

emolteni:
> > Hi, please consider that, since this morning, we have started making new changes in the yambo parser, with the aim of trying to let NOMAD plot _spectra_ from Yambo calculations, and therefore completely independent and unrelated wrt to the subject of our pull request you are examining (which is about reporting cell parameters, chemical composition etc). Should we make another fork, or what else, in order to continue our new (tentative) development for spectra, while you are examining our PR? Thanks.
> 
> Hi, I suggest you finalize this pr and then i merge it. Then you can rebase your branch and work on the new chenges. Again, please remove the QE parser stuff from this pr, either by creating a new branch and updating the source branch of this pr or simply not commiting your changes to the QE parser. Thanks!

Hi, we have just reverted the yambo parser in our fork to the version prior to this morning changes, and the QE parser to the version in the general NOMAD, and deleted the QE parser BACK.

emolteni:
Hi, is there anything else I should do on my part on this PR about the parsing of cell params etc from Yambo?  I see a message about conflicts in the QE parser: by clicking on "resolve conflicts" I see the following part highlighted:  

<<<<<<< develop
                    value = np.dot(
                        value.magnitude if hasattr(value, 'magnitude') else value,
                        cell.magnitude if hasattr(cell, 'magnitude') else cell
                    ) * cell.units if hasattr(cell, 'units') else 1.0
=======
                    value = (
                        np.dot(
                            value.magnitude if hasattr(value, 'magnitude') else value,
                            cell.magnitude if hasattr(cell, 'magnitude') else cell,
                        )
                        * cell.units
                        if hasattr(cell, 'units')
                        else 1.0
                    )
>>>>>>> develop

which is not a part we have written, and moreover it is not present either in the current https://github.com/nomad-coe/electronic-parsers/tree/develop/electronicparsers/quantumespresso or in our corresponding version of the QE parser (https://github.com/emolteni/electronic-parsers/tree/develop/electronicparsers/quantumespresso) which is just the standard NOMAD one as of last month.

electronicparsers/yambo/parser.py
```diff
@@ -51,7 +51,11 @@ def __init__(self):
     def init_quantities(self):
         re_f = r'[-+]*\d*\.\d+[Ee]*[-+]*\d*'
 
+
```
ladinesa: remove blank lines

electronicparsers/yambo/parser.py
```diff
@@ -159,6 +163,16 @@ def init_quantities(self):
             ),
         ]
 
+
```
ladinesa: remove commented lines

electronicparsers/yambo/parser.py
ladinesa: ah now i understand that you have the same quantity def for `qp_properties_quantity`. I suggest you define this once and use in both places.

electronicparsers/yambo/parser.py
```diff
@@ -475,9 +496,16 @@ def init_quantities(self):
                 repeats=True,
                 sub_parser=TextParser(quantities=module_quantities),
             ),
+            Quantity(
```
ladinesa: is this used anywhere?

electronicparsers/yambo/parser.py
```diff
@@ -710,23 +741,59 @@ def parse_input(self):
...
             n_atoms = self.netcdf_parser.N_ATOMS
             atom_numbers = np.hstack(
                 [
                     [self.netcdf_parser.atomic_numbers[int(n)]] * int(n_atoms[int(n)])
                     for n in range(len(n_atoms))
                 ]
             )
+
+#     we define the process_and_select function within the parse_input function
+            def process_and_select(positions, max_n_atoms, n_atoms):
```
ladinesa: defined this outside the if statement, please provide doc.

emolteni: It is indeed defined _inside_ the if statement.  To which "doc" (documentation?) are you referring?

ladinesa: please add a docstring to the function

ladinesa: consider changing function name to reflect that it really does

emolteni: We have just changed the function name to correct_position_array.

ladinesa:
can you perhaps remove the quantum espresso changes in this pr?

emolteni:
> can you perhaps remove the quantum espresso changes in this pr?

I have just verified my version of the QE parser is identical to the current one on nomad-coe...  .  
Moreover, Here https://github.com/nomad-coe/electronic-parsers/compare/develop...emolteni:electronic-parsers:develop 
it says "Able to merge. These branches can be automatically merged."

On the other hand, pages like this  https://github.com/nomad-coe/electronic-parsers/pull/286/files  are possibly comparing subsequent versions of "our" version of the QE parser, where we just updated it in order to keep it up to date with the changing qe parser in nomad-coe.

ladinesa:
yeah you are right, maybe rebase it to nomad:develop and it would go away.

emolteni:
> yeah you are right, maybe rebase it to nomad:develop and it would go away.

Is there a way to do a rebase here on github?

ladinesa:
> > yeah you are right, maybe rebase it to nomad:develop and it would go away.
> 
> Is there a way to do a rebase here on github?

I created a pr in your fork to rebase it.

emolteni:
> > > yeah you are right, maybe rebase it to nomad:develop and it would go away.
> > 
> > 
> > Is there a way to do a rebase here on github?
> 
> I created a pr in your fork to rebase it.

Thank you.

So, in  https://github.com/emolteni/electronic-parsers/pull/1,
in "Merge pull request"
I should choose "Rebase and merge"?

ladinesa:
you can also just squash so you do not see all commits

ladinesa:
I do not understand what happened there but can you perhaps repeat it? rebase it again, i was expecting that the other changes will vanish.

emolteni:
> I do not understand what happened there but can you perhaps repeat it? rebase it again, i was expecting that the other changes will vanish.

I have just made a rebase again. I also don't understand why it still shows the clickable link "rebase" when one has already rebased and changed nothing afterwards. And yes, you are right, it keeps showing the same list of changes to some of the parsers.

ladinesa:
can you do it locally then push the changes to your develop branch.

electronicparsers/yambo/metainfo/yambo.py
```diff
@@ -65,7 +65,7 @@ class x_yambo_io(MSection):
 
 class x_yambo_parameters(MSection):
     m_def = Section(validate=False)
-
+
```
ladinesa: run `ruff check --fix` there are some whitespaces here.

electronicparsers/yambo/parser.py
```diff
@@ -50,8 +50,9 @@ def __init__(self):
 
     def init_quantities(self):
         re_f = r'[-+]*\d*\.\d+[Ee]*[-+]*\d*'
-
+
```
ladinesa: also here

electronicparsers/yambo/parser.py
```diff
@@ -50,8 +50,9 @@ def __init__(self):
 
     def init_quantities(self):
         re_f = r'[-+]*\d*\.\d+[Ee]*[-+]*\d*'
-
+  
         io_quantities = [
+
```
ladinesa, line 53: also here

electronicparsers/yambo/parser.py
```diff
@@ -603,6 +606,9 @@ def parse_calculation(self, source, target=None):
             'valence_conduction',
             [source.get('valence', 0.0), source.get('conduction', 0.0)],
         )
+
```
ladinesa: also here

electronicparsers/yambo/parser.py
```diff
@@ -717,23 +723,62 @@ def parse_input(self):
...
+            max_n_atoms = self.netcdf_parser.get('MAX_ATOMS')
             n_atoms = self.netcdf_parser.N_ATOMS
             atom_numbers = np.hstack(
                 [
                     [self.netcdf_parser.atomic_numbers[int(n)]] * int(n_atoms[int(n)])
                     for n in range(len(n_atoms))
                 ]
             )
+
+            def correct_position_array(positions, max_n_atoms, n_atoms):
```
ladinesa: you do not need to define this function here. either you simply insert the code here or define a  class method 
```python
def get_positions(self):
    positions = self.netcdf_parser.get('ATOM_POS', [])
    ...
    return positions
````
if you run into a linting error with long functions. You do not need to pass any arg as these can be extracted from self.

File: electronicparsers/yambo/parser.py [hunk 21]
```diff
@@ -717,23 +723,62 @@ def parse_input(self):
...
+                to a chemical species, then we extract the first n_atoms 
+                (value of n_atoms for each chemical species present in the system)
+                from each block, and finally we reassemble the modified blocks
+                into to corrected positions array'''
+                positions = np.array(positions)
+                blocks = []
+                selected = []
+
+                positions = positions.reshape(-1, 3)
+
```
ladinesa: check unnecessary whitespaces, run linting check.

electronicparsers/yambo/parser.py
```diff
@@ -717,23 +723,62 @@ def parse_input(self):
...
                 labels=[chemical_symbols[int(n)] for n in atom_numbers],
             )
+##########################            
+
             if self.netcdf_parser.LATTICE_VECTORS is not None:
                 system.atoms.lattice_vectors = (
                     self.netcdf_parser.LATTICE_VECTORS * ureg.bohr
                 )
 
+
```
ladinesa: here also

electronicparsers/yambo/parser.py
```diff
@@ -885,7 +930,7 @@ def parse(self, filepath, archive, logger):
                 key = key.strip().replace('/', '').replace(' ', '_').lower()
                 val = val == 'yes' if val in ['yes', 'no'] else val
                 setattr(run, 'x_yambo_%s' % key, val)
-
+
```
ladinesa: here also

tests/test_yamboparser.py
```diff
@@ -216,3 +216,17 @@ def test_4(parser):
...
+        'tests/data/yambo/CH4_db_minimal/r_setup', archive, None
+    )
+    run = archive.run[-1]
+    system = run.system
+    assert run.x_yambo_input.x_yambo_file == './SAVE//ns.db1'
+    assert system[0].atoms.positions[0][0] == approx(0.67500)
+    assert x_yambo_parameters[0].x_yambo_max_atoms_species[0] == 4
+    assert x_yambo_parameters[0].x_yambo_n_atom_species[0] == 4
+    assert x_yambo_parameters[0].x_yambo_n_atom_species[1] == 1
+
```
ladinesa: here also

ladinesa:
Hi @emolteni I just found some minor issues with linting and a suggestion on defining `get_positions`. If you manage to address these, we can merge.

electronicparsers/yambo/parser.py
emolteni: Sorry, but it's not so clear to me what you mean by "simply insert the code here": you mean just letting that part of code there without making it a function?   About the other alternative:  you mean putting all the rest of our current version of that function after "positions = self.netcdf_parser.get('ATOM_POS', [])" ?  But "positions" is already defined as "positions = self.netcdf_parser.get('ATOM_POS', [])" ....so maybe I am missing something.   By the way, if I remember well, we indeed needed to treat  positions as an array, in order to split it into blocks etc.

ladinesa: yes you do not need to define the function as you actually do not need to pass any argument as these are all defined at the top of the function, also you are not reusing it. If you want to make it cleaner you can define it as `def get_positions(self)` and copy everything from `positions = self.netcdf_parser.get('ATOM_POS', [])` until `return positions` and do `positions = self.get_positions()`

emolteni: Ok, thanks for your explanations.  I am going to try either of the two options tomorrow, locally, and test if my modified version still works on my data (to be sure I have implemented them correctly), then I will update here on github.

emolteni:
Hi @ladinesa, I think I have addressed all your observations: I have modified the part on correcting the positions array in the first of the 2 ways you suggested (i.e. without defining that part as a function)  and checked that it still works on my files, and I removed the unnecessary empty lines and whitespaces in the parser, metainfo and test file.

electronicparsers/yambo/parser.py
```diff
@@ -717,16 +717,45 @@ def parse_input(self):
...
+            max_n_atoms = self.netcdf_parser.get('MAX_ATOMS')
             n_atoms = self.netcdf_parser.N_ATOMS
             atom_numbers = np.hstack(
                 [
                     [self.netcdf_parser.atomic_numbers[int(n)]] * int(n_atoms[int(n)])
                     for n in range(len(n_atoms))
                 ]
             )
+
+            '''We split the positions array into blocks, each corresponding
```
ladinesa: make this an ordinary comment, i.e use `#` not `'''`

electronicparsers/yambo/parser.py
```diff
@@ -717,16 +717,45 @@ def parse_input(self):
...
+            n_blocks = int( int(n_points) // int(max_n_atoms) )
+
+            for i in range(n_blocks):
+                start_idx = int(i) * int(max_n_atoms)
+                end_idx = (int(i) + 1) * int(max_n_atoms)
+                block = positions[start_idx:end_idx]
+                blocks.append(block)
+
+
+            for i, block in enumerate(blocks):
```
ladinesa: you do not need to split this with the loop in l741, i.e. do not define blocks just put l749-752 in l745

electronicparsers/yambo/parser.py
```diff
@@ -717,16 +717,45 @@ def parse_input(self):
...
+            to a chemical species, then we extract the first n_atoms
+            (value of n_atoms for each chemical species present in the system)
+            from each block, and finally we reassemble the modified blocks
+            into to corrected positions array'''
+            positions = np.array(positions)
+            blocks = []
+            selected = []
+            positions = positions.reshape(-1, 3)
+            n_points = positions.shape[0]
+            n_blocks = int( int(n_points) // int(max_n_atoms) )
```
ladinesa: there is a numpy deprecation warning in l739 l 742 and l743 can you please address these?

electronicparsers/yambo/parser.py
emolteni: Done.

electronicparsers/yambo/parser.py
emolteni: Done

electronicparsers/yambo/parser.py
emolteni: I wonder whether the warning is related to the use of "int" (I have seen past reports on the syntax of "int" having changed in numpy...), but there are other instances of "int" in the std part of the parser....  I have then tried to do "run and debug" of my parser on a codespace, but I find an error on importing netCDF4, that I find also if doing the same on the standard version of the parser, therefore there must be some command I am missing.  Are there other ways you can suggest for me to test this kind of issue?  (if it were an error, it would of course be easier to test it, but it is just a warning, and in fact the parser works fine on my data).

ladinesa: maybe do positions.shape.tolist()

emolteni: I have tried (locally on my pc, not in my github branch):  
(.....)
positions = positions.reshape(-1, 3)
pos_shape_list = positions.shape.tolist()
n_points = pos_shape_list[0]

...and the rest unchanged.  

But, when trying it on my Yambo files (with:  uv run nomad parse ....)  I get the error:
pos_shape_list = positions.shape.tolist()
                     ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'tuple' object has no attribute 'tolist'

The "tolist" method should convert an array into a nested list:  positions.shape is (has to be) an array, in particular a vector, containing 2 elements = the number of rows  ( = total number of lines in the total coordinate list) and number of columns (3) of the "positions" array.   So I wonder what the problem is.

emolteni: Ok, well, from our "human" point of view, positions.shape should be a vector (that is a particular case of an array... I hope I'm right on this at least), but in fact in python it turns out to be a tuple, and that's why we cannot apply "tolist" to it.  I have just read some documentation and learned that a tuple and a vector are in python much less similar objects than one would expect naively from those two words.

ladinesa: sorry that was a mistake, indeed shape produces a tuple, not sure then about the error, try looking at the type of the variables which you are converting.

emolteni: Ok. So, going back to the numpy deprecation warning you mentioned (for my version of the parser, without the use of "tolist"), can you tell me whether there were more details in the warning message, which may suggest the kind of problem? (whether it is about the use of "int", or the division //  .....etc).  I have done several different tries, but so far I have not been able to find a way to run the parser (either locally on my pc or on Jupyterlab) letting it show deprecation warnings. So I am not even able to check the effect of my possible test changes on those deprecation warnings. Thanks.

ladinesa: Please look at the workflow.actipns log.

emolteni: I have fixed the problem in our test_5 in test_yamboparser.py, putting the correct numerical value of system[0].atoms.positions[0][0].magnitude (and also inserting "magnitude", which I noticed is present in all the other tests about non-integer numerical values), which is indeed 0.675 Angstrom, i.e. 6.7500e-11 (in meters).  I have also seen the explicit deprecation warning message for l739,l742,l743:  it complains about the conversion of an array with ndim > 0 to a scalar.  The variable that appears in all these 3 lines is max_n_atoms: this quantity is defined few lines above thus: max_n_atoms = self.netcdf_parser.get('MAX_ATOMS').   Maybe I should make max_n_atoms  explicitly a scalar, or take its first value (which I assume is the only one...)? The other possible problematic quantity is n_blocks = int( int(n_points) // int(max_n_atoms) ),  where n_points = positions.shape[0] so I would say this is a scalar.

ladinesa: please convert to the units that is used in the output file you compare to the number that you parse i.e. `positions.to('angstrom').magnitude == approx(...)`

ladinesa: sorry I did not have time to look at this issue in detail but please do the necessary fix to get rid of this warning, maybe look at this value for `MAX_ATOMS` and if indeed it is a a list then do the appropriate fix. Thanks.

emolteni: I have just addressed the last 2 points:
1)   in the parser, I am now using  max_n_atoms_1 = max_n_atoms[0]   instead of max_n_atoms  (indeed  max_n_atoms, although containing 1 element only, is of type numpy.array (with shape = (1,) ), and the deprecation warning indeed suggested extracting 1 element of an array):  this new version of the parser still works correctly on my data
2)  I have put  positions.to('angstrom')  in test_5 in the test_yamboparser.py

electronicparsers/yambo/parser.py
```diff
@@ -717,16 +717,41 @@ def parse_input(self):
             system = System()
             run.system.append(system)
             positions = self.netcdf_parser.get('ATOM_POS', [])
+            max_n_atoms = self.netcdf_parser.get('MAX_ATOMS')
+            max_n_atoms_1 = max_n_atoms[0]
```
ladinesa: you can simply put `max_n_atoms = self.netcdf_parser.get('MAX_ATOMS')[0]`

electronicparsers/yambo/parser.py
```diff
@@ -717,16 +717,41 @@ def parse_input(self):
...
+                start_idx = int(i) * int(max_n_atoms_1)
+                end_idx = (int(i) + 1) * int(max_n_atoms_1)
+                block = positions[start_idx:end_idx]
+                n_to_select = int(n_atoms[int(i)])
+                selected_from_block = block[:n_to_select]
+                for point in selected_from_block:
+                    selected.append(point)
+
+            positions=np.array(selected)
+
```
ladinesa, line 751: remove linebreak

ladinesa:
@emolteni thanks for your patience I think we can merge this now after you resolve the last couple of minor tweaks

electronicparsers/yambo/parser.py
emolteni: Done

electronicparsers/yambo/parser.py
emolteni: Done

emolteni:
> @emolteni thanks for your patience I think we can merge this now after you resolve the last couple of minor tweaks

I have just addressed both your last 2 suggestions.  Thanks for your suggestions and supervising.

ladinesa:
Test still failing.

emolteni:
> Test still failing.

Hi, can you remind me how (in case there is a way) I can run a parser test locally on my Oasis?  I have started looking at this last error on test_5 of the test_yamboparser.py, and I am a bit confused about the use of the x_yambo.... variables, the related syntax, the way they are defined in the metainfo etc. So I would like to make some tries and check locally if the modified test_yamboparser.py works.  E.g.  I think we may have to assert  something like  x_yambo_input.x_yambo_parameters[0].x_yambo_max_atoms_species[0]   instead of  simply  x_yambo_parameters[0].x_yambo_max_atoms_species[0]

ladinesa:
> > Test still failing.
> 
> Hi, can you remind me how (in case there is a way) I can run a parser test locally on my Oasis? I have started looking at this last error on test_5 of the test_yamboparser.py, and I am a bit confused about the use of the x_yambo.... variables, the related syntax, the way they are defined in the metainfo etc. So I would like to make some tries and check locally if the modified test_yamboparser.py works. E.g. I think we may have to assert something like x_yambo_input.x_yambo_parameters[0].x_yambo_max_atoms_species[0] instead of simply x_yambo_parameters[0].x_yambo_max_atoms_species[0]

you have to do `pytest -sxv tests/test_yambo_parser.py`. The fix is it shoul be `run.x_yambo_input.x_yambo_parameters[0]....`

emolteni:
> > > Test still failing.
> > 
> > 
> > Hi, can you remind me how (in case there is a way) I can run a parser test locally on my Oasis? I have started looking at this last error on test_5 of the test_yamboparser.py, and I am a bit confused about the use of the x_yambo.... variables, the related syntax, the way they are defined in the metainfo etc. So I would like to make some tries and check locally if the modified test_yamboparser.py works. E.g. I think we may have to assert something like x_yambo_input.x_yambo_parameters[0].x_yambo_max_atoms_species[0] instead of simply x_yambo_parameters[0].x_yambo_max_atoms_species[0]
> 
> you have to do `pytest -sxv tests/test_yambo_parser.py`. The fix is it shoul be `run.x_yambo_input.x_yambo_parameters[0]....`


I have tried it locally:   assert run.x_yambo_input.x_yambo_parameters[0].x_yambo_max_atoms_species[0] == 4
yields  "KeyError: 0".  I see that "Key error" may occur when one applies to a dictionary a function which has to act on a list. Indeed I checked that   run.x_yambo_input.x_yambo_parameters   is a dictionary.   Moreover, in the other cases in test_yamboparser.py,  the used syntax is  x_yambo_parameters['<some string>'],  (some text string instead of an integer), and in test_2 (similar to our case:  an r* file and a ns.db1 file)  we have:  
assert run.x_yambo_input.x_yambo_parameters['Electrons'] == approx(3.0)    
where "Electrons" is the "key" corresponding to the value 3.0  _in the  r*  file,  not in the ns.db1._  

By putting:  assert run.x_yambo_input.x_yambo_parameters['Max atoms/species'] == 4    (and commenting the 2 following "assert" statements)     the test runs correctly ('Max atoms/species'  is the key in our r* file of test_5  corresponding to value 4 and, importantly, corresponding to  max_n_atoms  in the parser, and to the 'MAX_ATOMS' string in our ns.dd1).

So I can modify our 2 further "assert" statements accordingly.

ladinesa:
> > > > Test still failing.
> > > 
> > > 
> > > Hi, can you remind me how (in case there is a way) I can run a parser test locally on my Oasis? I have started looking at this last error on test_5 of the test_yamboparser.py, and I am a bit confused about the use of the x_yambo.... variables, the related syntax, the way they are defined in the metainfo etc. So I would like to make some tries and check locally if the modified test_yamboparser.py works. E.g. I think we may have to assert something like x_yambo_input.x_yambo_parameters[0].x_yambo_max_atoms_species[0] instead of simply x_yambo_parameters[0].x_yambo_max_atoms_species[0]
> > 
> > 
> > you have to do `pytest -sxv tests/test_yambo_parser.py`. The fix is it shoul be `run.x_yambo_input.x_yambo_parameters[0]....`
> 
> I have tried it locally: assert run.x_yambo_input.x_yambo_parameters[0].x_yambo_max_atoms_species[0] == 4 yields "KeyError: 0". I see that "Key error" may occur when one applies to a dictionary a function which has to act on a list. Indeed I checked that run.x_yambo_input.x_yambo_parameters is a dictionary. Moreover, in the other cases in test_yamboparser.py, the used syntax is x_yambo_parameters[''], (some text string instead of an integer), and in test_2 (similar to our case: an r* file and a ns.db1 file) we have: assert run.x_yambo_input.x_yambo_parameters['Electrons'] == approx(3.0) where "Electrons" is the "key" corresponding to the value 3.0 _in the r* file, not in the ns.db1._
> 
> By putting: assert run.x_yambo_input.x_yambo_parameters['Max atoms/species'] == 4 (and commenting the 2 following "assert" statements) the test runs correctly ('Max atoms/species' is the key in our r* file of test_5 corresponding to value 4 and, importantly, corresponding to max_n_atoms in the parser, and to the 'MAX_ATOMS' string in our ns.dd1).
> 
> So I can modify our 2 further "assert" statements accordingly.

o sorry no index as it is a json object.

emolteni:
> > > > > Test still failing.
> > > > 
> > > > 
> > > > Hi, can you remind me how (in case there is a way) I can run a parser test locally on my Oasis? I have started looking at this last error on test_5 of the test_yamboparser.py, and I am a bit confused about the use of the x_yambo.... variables, the related syntax, the way they are defined in the metainfo etc. So I would like to make some tries and check locally if the modified test_yamboparser.py works. E.g. I think we may have to assert something like x_yambo_input.x_yambo_parameters[0].x_yambo_max_atoms_species[0] instead of simply x_yambo_parameters[0].x_yambo_max_atoms_species[0]
> > > 
> > > 
> > > you have to do `pytest -sxv tests/test_yambo_parser.py`. The fix is it shoul be `run.x_yambo_input.x_yambo_parameters[0]....`
> > 
> > 
> > I have tried it locally: assert run.x_yambo_input.x_yambo_parameters[0].x_yambo_max_atoms_species[0] == 4 yields "KeyError: 0". I see that "Key error" may occur when one applies to a dictionary a function which has to act on a list. Indeed I checked that run.x_yambo_input.x_yambo_parameters is a dictionary. Moreover, in the other cases in test_yamboparser.py, the used syntax is x_yambo_parameters[''], (some text string instead of an integer), and in test_2 (similar to our case: an r* file and a ns.db1 file) we have: assert run.x_yambo_input.x_yambo_parameters['Electrons'] == approx(3.0) where "Electrons" is the "key" corresponding to the value 3.0 _in the r* file, not in the ns.db1._
> > By putting: assert run.x_yambo_input.x_yambo_parameters['Max atoms/species'] == 4 (and commenting the 2 following "assert" statements) the test runs correctly ('Max atoms/species' is the key in our r* file of test_5 corresponding to value 4 and, importantly, corresponding to max_n_atoms in the parser, and to the 'MAX_ATOMS' string in our ns.dd1).
> > So I can modify our 2 further "assert" statements accordingly.
> 
> o sorry no index as it is a json object.


These two work correctly:
   assert run.x_yambo_input.x_yambo_parameters['Max atoms/species'] == 4
    assert run.x_yambo_input.x_yambo_parameters['No. of atom species'] == 2

while this one yields  KeyError: 'number of H  atoms':
assert run.x_yambo_input.x_yambo_parameters['number of H  atoms'] == 4
(which is meant to correspond to   assert x_yambo_parameters[0].x_yambo_n_atom_species[0] == 4)

A difference is that 'number of H  atoms'   is in the section "[02.01] Unit cells"  of our r* file, while 'Max atoms/species' and 'No. of atom species' are in section "[RD./SAVE//ns.db1]" of the same file.

ladinesa:
please make the necessary assertions in the correct sub-sections to test out the added functionality. I think you did not make any changes to the input_parameters? so these should actually not be added to the test only the positions part.

emolteni:
> please make the necessary assertions in the correct sub-sections to test out the added functionality. I think you did not make any changes to the input_parameters? so these should actually not be added to the test only the positions part.

It is true that, in our version, we have _modified_ (modified, not added) the definition of positions, but we also have added the _max_n_atoms_ variable ( max n atoms / species), which we use to split the initially wrong (n of lines: max_n_atoms times the number of different chemical species in the system) list of coordinates into blocks and then to take only the first n_atoms (value of n_atoms for a given species) lines of coordinates for each of the chemical species in the system. That's why we would like to test that NOMAD now correctly parses both the max n atoms / species  and the actual (n_atoms) numbers of atoms for each of the species present in our system.

emolteni:
> > please make the necessary assertions in the correct sub-sections to test out the added functionality. I think you did not make any changes to the input_parameters? so these should actually not be added to the test only the positions part.
> 
> It is true that, in our version, we have _modified_ (modified, not added) the definition of positions, but we also have added the _max_n_atoms_ variable ( max n atoms / species), which we use to split the initially wrong (n of lines: max_n_atoms times the number of different chemical species in the system) list of coordinates into blocks and then to take only the first n_atoms (value of n_atoms for a given species) lines of coordinates for each of the chemical species in the system. That's why we would like to test that NOMAD now correctly parses both the max n atoms / species and the actual (n_atoms) numbers of atoms for each of the species present in our system.


Hi, sorry for the delay.  I have been busy with different subjects.  

So: in our "test_5" in the test_yamboparser.py, these 2 assertions work correctly:
assert run.x_yambo_input.x_yambo_parameters['Max atoms/species'] == 4
assert run.x_yambo_input.x_yambo_parameters['No. of atom species'] == 2

while these 2 do not:
assert run.x_yambo_input.x_yambo_parameters['number of H atoms'] == 4
assert run.x_yambo_input.x_yambo_parameters['number of C  atoms'] == 1

As I mentioned before, a relevant difference that occurs to me is that  'number of H atoms' and  'number of C  atoms' are (together) in a different section of our test's mainfile (https://github.com/emolteni/electronic-parsers/tree/develop/tests/data/yambo/CH4_db_minimal/r_setup), i.e. in the section "[02.01] Unit cells", at a difference with  'Max atoms/species' and 'No. of atom species', which are in the section "[RD./SAVE//ns.db1]". 

I have noticed that, in the other tests contained in test_yamboparser.py, sometimes we have calc[0].<somequantity>, calc[1].<somequantity>,  or method[0].<somequantity>, method[1].<somequantity>, referring to quantities defined in _different sections_ of the mainfile. 


So I wonder whether we can do the same for the quantities for our assertions in test_5. 
However, these are quantities under "run", not under some subsection (calc, method, system..) of run, and I have not found - even in the test_xxx.py files for other codes - instances of run[1], run[2] etc, but only, in some cases, run[0].

On the other hand, in  test_yamboparser.py  we always have  "run = archive.run[-1]",  which should mean "the last element of  archive.run", am I correct?

I have tried to define "run = archive.run" (instead of "run = archive.run[-1]") in test_5, with   system = run[-1].system  (instead of  system = run.system)  and  then referring to specific "elements" of "run", found in different sections of the mainfile, similarly to what is done for quantities in method or calc, i.e. using run[0] or run[1], run[-1], run[-2] each followed by the name of the quantity etc. 

Now, these:
assert run[-1].x_yambo_input.x_yambo_parameters['Max atoms/species'] == 4
assert run[-1].x_yambo_input.x_yambo_parameters['No. of atom species'] == 2
do work, and indeed they are equivalent to their standard definition...

while, for 'number of H  atoms'  or  'number of C  atoms':
using  "assert run[0].x_yambo_input.x_yambo_parameters['number of H  atoms'] == 4"  or the equivalent with run[-1]  yields  "KeyError: 'number of H  atoms'";
using  run[1]  or  run[-2]  yields  "IndexError: list index out of range";
using   "assert run.x_yambo_input.x_yambo_parameters['number of H  atoms'] == 4"  yields  "AttributeError: 'MSubSectionList' object has no attribute 'x_yambo_input'"

...so it is not clear to me what is going on here.   
Is "run" composed of several elements, such as calc or method seem to be?  (with different elements or components, [1], [2], ... referring to quantities contained in different sections of the mainfile) 
Why in all the (other) tests in  test_yamboparser.py  are we always using   run = archive.run[-1]   which I assume should indicate the last element of run? What about other elements?

emolteni:
Hi, in the end I have found a way to make our test_5 work (in test_yamboparser.py), at the same time using assertions involving the part we have contributed to the parser, i.e. the correct handling of the number of atoms per chemical species etc.  I have just committed the mentioned changes in test_yamboparser.py in my fork, and I have locally verified the tests work correctly, without errors. Just for completeness, I am also reporting here the updated form of the last lines in test_5:

    assert run.x_yambo_input.x_yambo_parameters['Max atoms/species'] == 4
    assert run.x_yambo_input.x_yambo_parameters['No. of atom species'] == 2
    assert system[0].atoms.positions.shape == (5, 3)

electronicparsers/yambo/parser.py
```diff
@@ -717,16 +717,38 @@ def parse_input(self):
             system = System()
             run.system.append(system)
             positions = self.netcdf_parser.get('ATOM_POS', [])
+            max_n_atoms = self.netcdf_parser.get('MAX_ATOMS')[0]
```
ladinesa: put a check here for max_n_atoms, if None will return e.g.
```max_n_atoms = int(self.netcdf_parser.get('MAX_ATOMS', [0])[0])
if not max_n_atoms:
 return
```
then you also do not need the integer conversions below

ladinesa:
I think this is ok to merge after you address the last comment.

electronicparsers/yambo/parser.py
emolteni: Thanks a lot.  I have inserted the check on max_n_atoms you suggested, and removed some of the integer conversions. Not all of them, since some of them, when removed, produced an error about objects that should be integers but could not be interpreted as such. I have tried the updated version on my test folder CH4_minimal and it works.

emolteni:
> I think this is ok to merge after you address the last comment.

Do you think you can merge it now, after my last changes?  Thanks.

electronicparsers/yambo/parser.py
ladinesa: hi sorry pls put in l720 max_n_atoms = self.netcdf_parser.get('MAX_ATOMS', [0])[0]`. Then, resolve all and merge.

emolteni: > hi sorry pls put in l720 max_n_atoms = self.netcdf_parser.get('MAX_ATOMS', [0])[0]`. Then, resolve all and merge.

I have performed this change, too.  I thought the merge was going to be done by you, as team member reviewing this PR, isn't it so?   (as explained e.g. here  https://nomad-lab.eu/prod/v1/docs/howto/develop/contrib.html#pull-requests-pr-github)

emolteni:
> I think this is ok to merge after you address the last comment.

As mentioned in the "resolved" conversation above, I have performed the last change you have suggested, i.e.     
max_n_atoms = self.netcdf_parser.get('MAX_ATOMS',[0])[0]

As for merging, I thought the merge was going to be performed by you, as team member reviewing this PR, isn't it so? (as explained e.g. here https://nomad-lab.eu/prod/v1/docs/howto/develop/contrib.html#pull-requests-pr-github)

PR #288, opened 2025-05-05
Implement qe workflow linking

ladinesa:


electronicparsers/quantumespresso/parser.py
```diff
@@ -2797,6 +2801,44 @@ def __init__(self):
...
+        self._child_archives = {}
+
+    def get_mainfile_keys(self, **kwargs):
+        """
+        Return the QE workflow files in the directory or sub-directory.
+        """
+        qe_files = []
+        filename = kwargs.get('filename')
+
+        max_dirs = 5
```
ndaelman-hu, line 2814: shouldn't this be a parameter set by the entrypoint?

electronicparsers/quantumespresso/parser.py
```diff
@@ -3612,3 +3659,49 @@ def parse(self, filepath, archive, logger):
...
+                for search_id in search_ids:
+                    for qe_file, workflow_archive in self._child_archives.items():
+                        if search_id.get('mainfile') in qe_file:
+                            # link qe workflow entry
+                            workflow_archive.workflow2 = SimulationWorkflow()
+                            workflow_archive.workflow2.inputs = archive.workflow2.inputs
+                            entry_archive = archive.m_context.load_archive(
+                                search_id.get('entry_id'), upload_id, None
+                            )
+                            if entry_archive.workflow2 is None:
```
ndaelman-hu: only the xspectra module sets `workflow2`, so these will have to be extended then

electronicparsers/quantumespresso/parser.py
```diff
@@ -3612,3 +3659,49 @@ def parse(self, filepath, archive, logger):
...
+                                continue
+                            workflow_archive.workflow2.outputs = (
+                                entry_archive.workflow2.outputs
+                            )
+                            workflow_archive.workflow2.tasks.extend(
+                                [
+                                    TaskReference(
+                                        name=f'{archive.run[0].program.name} calculation',
+                                        task=archive.workflow2,
+                                        inputs=workflow_archive.workflow2.inputs,
```
ndaelman-hu: Aren't these inputs already covered at the global workflow entry?

electronicparsers/quantumespresso/parser.py
```diff
@@ -3612,3 +3659,49 @@ def parse(self, filepath, archive, logger):
...
+                                    ),
+                                    TaskReference(
+                                        name=f'{entry_archive.run[0].program.name} calculation',
+                                        task=entry_archive.workflow2,
+                                        inputs=[Link(section=archive.workflow2)],
+                                        outputs=workflow_archive.workflow2.outputs,
```
ndaelman-hu: Aren't these outputs already covered at the global workflow entry?

electronicparsers/quantumespresso/parser.py
```diff
@@ -3612,3 +3659,49 @@ def parse(self, filepath, archive, logger):
...
+                search_ids = search(
+                    owner='visible',
+                    user_id=archive.metadata.main_author.user_id,
+                    query={'upload_id': upload_id},
+                    required=MetadataRequired(include=['entry_id', 'mainfile']),
+                ).data
+
+                for search_id in search_ids:
+                    for qe_file, workflow_archive in self._child_archives.items():
+                        if search_id.get('mainfile') in qe_file:
```
ndaelman-hu: I'm not following this line:

1. Doesn't `get_mainfile_keys` produce keys based on the module name (except for "pwscf") instead of the file names?
2. Why do we need a NOMAD query targeting our own entry? Can't we get "entry_id" from `archive`?

electronicparsers/quantumespresso/parser.py
```diff
@@ -3612,3 +3659,49 @@ def parse(self, filepath, archive, logger):
...
+                                    TaskReference(
+                                        name=f'{entry_archive.run[0].program.name} calculation',
+                                        task=entry_archive.workflow2,
+                                        inputs=[Link(section=archive.workflow2)],
+                                        outputs=workflow_archive.workflow2.outputs,
+                                    ),
+                                ]
+                            )
+            except Exception:
+                self.logger.error('Error linking workflow entries.')
```
ndaelman-hu: `Error retrieving task entries or linking them.`

You could also split the error messages by code block.

electronicparsers/quantumespresso/parser.py
```diff
@@ -2797,6 +2801,44 @@ def __init__(self):
...
+                    f
+                    for f in glob(f'{os.path.dirname(filename)}/{pattern}')
+                    if f != filename
+                ]
+            )
+            if qe_files:
+                break
+            pattern = os.path.join('**', pattern)
+
+        # key for the overall workflow
```
ndaelman-hu: I guess you mean "generate keys for all relevant workflow tasks"

electronicparsers/quantumespresso/parser.py
ndaelman-hu, line 1: Just some minor comments and 1 conceptual one.

This setup indicates that the bands parser will itself have to access the previous workflow step to populate some necessary band structure fields. Atm, this is done via `nomad.search`, but feel free to raise alternatives.

electronicparsers/__init__.py
ndaelman-hu, line 1: Where can we tell users that these workflows will only be picked up if they're in the same upload?
This might not be obvious to some, as these are separated over several git repos.

electronicparsers/quantumespresso/parser.py
ladinesa: no, i think is not too significant

electronicparsers/quantumespresso/parser.py
ladinesa: yes, you are right, i will add workflow2 creation in the qe parsers.

electronicparsers/quantumespresso/parser.py
ladinesa: yes, indeed TaskReference normalizes also the inputs and outputs

electronicparsers/quantumespresso/parser.py
ladinesa: no, i return the filename, i will try the context method `resolve_archive_url` for this

electronicparsers/quantumespresso/parser.py
ladinesa: this could be problematic as i set parser level higher for pwscf.

electronicparsers/__init__.py
ladinesa: this is an automatic workflow generation so maybe it is implied that it is restricted to the upload, i can put some comment in the qe parser

electronicparsers/quantumespresso/parser.py
ladinesa: https://github.com/nomad-coe/workflow-parsers/pull/56

electronicparsers/quantumespresso/parser.py
ladinesa: True

electronicparsers/quantumespresso/parser.py
ladinesa: seem to work, removing search

electronicparsers/quantumespresso/parser.py
ndaelman-hu: Then I'll add the combining logic here, after your PR.

electronicparsers/__init__.py
ndaelman-hu: Maybe in the QE parser's README?

ndaelman-hu:
@ladinesa Did you test whether this works for the current QE workflows?

electronicparsers/quantumespresso/parser.py
ndaelman-hu: I think that this approach should be documented somewhere. Do the new parsers have docs? Else I add to those of the schema.

ladinesa:
> @ladinesa Did you test whether this works for the current QE workflows?

yes

electronicparsers/quantumespresso/parser.py
ladinesa: not that i know of

ndaelman-hu (approved):

ndaelman-hu:
> > @ladinesa Did you test whether this works for the current QE workflows?
> 
> yes

In general, I think having any screenshots, especially as context for those reading this PR in the future would be useful.

ladinesa:
image

PR #293, opened 2025-06-17
basic rspt parser

JFRudzinski:
This is basically a shell parser so that some users can upload DFT and DMFT data from the RSPt software and have minimal search capabilities, etc. They will then work on developing a "real" parser for their code which can later be used for reprocessing


JFRudzinski:
@ladinesa This is meant to be a quick solution for some users who belong to a large collaborative project. I set this up so that they can meet some deadline for a report/publication, but will then set them up on a route to creating a proper plugin with the new schema. 

I am sure that my approach here is hacky, but also I do not want to spend too much more time on it, so if you can identify the essentials to make it acceptable in your opinion, that would be great.

I am waiting for some feedback from my data steward contact, but I think we can mostly still finalize in the mean time.

electronicparsers/rspt/parser.py
```diff
@@ -0,0 +1,141 @@
...
+    def __init__(self):
+        super().__init__()
+        self._method_parser = MethodParser()
+
+        self._parser = BasicParser(
+            'RSPt',
+            program_version=r'RSPt version number:\s*rspt\.(\d+\.\d+)',
+        )
+
+    def get_cif_files(self, auxillary_files):
```
ladinesa: no need for this function as used only once? you can just do maybe `cif_files  = [f for fin in aux_files if f.endswith('.cif')]`

electronicparsers/rspt/parser.py
```diff
@@ -0,0 +1,141 @@
...
+            logger.warning(
+                'RSPtParser method not recognized, or multiple distinct methods found.'
+            )
+
+        # PARSE SYSTEM
+        cif_files = self.get_cif_files(self._auxillary_files)
+        if cif_files:
+            if len(cif_files) > 1:
+                logger.warning(
+                    f'RSPtParser found multiple CIF files, using {cif_files[0]}.'
```
ladinesa: remove formatted text

ladinesa (approved):

ladinesa:
Maybe add a test data, just maybe parse it in test  no need to assert any quantity.

JFRudzinski:
> Maybe add a test data, just maybe parse it in test no need to assert any quantity.

Question about the logger for tests:

If I leave my logger statements as they are, i.e., using the input logger from the parse() function, then the logger output shows up in the GUI when I launch locally, but when I run pytest the logger is None.

On the other hand, if I instead set `self.logger = get_logger(__name__)` in the parser init() function and use that instead, then the pytest logs properly, but the logger output does not show up in the GUI when I launch locally.

Do you understand this? What is the correct approach?

ladinesa:
in `parse`, `logger = logger if logger is not None else get_logger(__name__)`

PR #299, opened 2025-08-04
Update numpy

blueraft:


ladinesa:
will you be able to fix the tests?

tests/dos_integrator.py
```diff
@@ -17,6 +17,10 @@
 #
 
 import numpy as np
+from importlib.metadata import version
+
+numpy_version = version('numpy')
+use_trapz = int(numpy_version.split('.')[0]) < 2
```
ladinesa: i do not think this is necessary as we only use updated numpy

blueraft: `nomad-lab` is still restricted to `numpy<2` so until that restriction is gone this would be necessary. But when that happens, this package would not require any further changes.

ladinesa: ah yes you are right. sorry

ladinesa: maybe put a TODO we do not forget to remove it

ladinesa (approved):

PR #308, opened 2025-09-29
Elk eigs quick fix

ndaelman-hu:
Address the issue of missing eigenvalues, as reported [here](https://matsci.org/t/elk-parser-fail/65516).

### DIFF
--- a/electronicparsers/elk/parser.py
+++ b/electronicparsers/elk/parser.py
@@ -463,7 +463,7 @@ class ElkParser:
         sec_eigenvalue = BandEnergies()
         sec_calc.eigenvalues.append(sec_eigenvalue)
         sec_eigenvalue.kpoints = self.eigenval_parser.get('kpoint')
-        eigs_occs = self.eigenval_parser.get('eigenvalue_occupancy', [])
+        eigenvalues_occupancies = self.eigenval_parser.get('eigenvalue_occupancy', [])
         n_spin = (
             1
             if self.mainfile_parser.get('spin_treatment', '').lower()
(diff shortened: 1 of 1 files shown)


ndaelman-hu:
@ladinesa this requires expedited deployment into the central NOMAD. Thx.

electronicparsers/elk/parser.py
```diff
@@ -463,22 +463,34 @@ def parse(self, filepath, archive, logger):
...
+            if "'NoneType' object cannot be interpreted as an integer" in str(error):
+                self.logger.warning(
+                    'Could not reshape eigenvalues/occupancies, setting to None'
+                )
+            else:
+                self.logger.warning(
+                    f'TypeError occurred during eigenvalue processing: {error}'
+                )
+            sec_eigenvalue.energies = None
+            sec_eigenvalue.occupancies = None
```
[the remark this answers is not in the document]
ndaelman-hu: In that case, I'd rather revert to the previous commit.

electronicparsers/elk/parser.py
```diff
@@ -463,22 +463,34 @@ def parse(self, filepath, archive, logger):
...
+        try:
+            eigenvalues_occupancies = np.reshape(
+                eigenvalues_occupancies,
+                (self.eigenval_parser.n_kpoints, n_spin, self.eigenval_parser.n_states, 3),
+            )
+            eigenvalues_occupancies = np.transpose(eigenvalues_occupancies, axes=(3, 1, 0, 2))
+            # first column is state index
+            sec_eigenvalue.energies = eigenvalues_occupancies[1] * ureg.hartree
+            sec_eigenvalue.occupancies = eigenvalues_occupancies[2]
+        except TypeError as error:
```
ladinesa: catch all exceptions, no need to determine specific error

electronicparsers/elk/parser.py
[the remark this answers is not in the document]
ladinesa: yes i agree just put a check that it is an array and of proper size instead of the try

electronicparsers/elk/parser.py
ndaelman-hu: Do you want both safety guards (check before + try-catch)?

ladinesa: just do the check before

PR #325, opened 2026-01-26
Openmx pseudo and corehole improvements

ondracka:
Few improvements to OpenMX parser, mostly fixes for an unpublished upload  kWr1F_ovSVe7YG6wrDpgVQ that now fails badly with 184 failures out of 186 OpenMX entries:

  - fix pseudopotential parsing
  - handle the empty atoms (no pseudo there and 0 atom number)
  - the atom label doesn't need to follow the standard atomic symbols (and OpenMX does not guarantee it does)
  - fix hardcoded 19 in the pseudo version
  - properly parse the code-specific atom labels
  - fix broken corehole parsing
  - add tests for all of this
  
  Nominating @ndaelman-hu for review as an original author of most of the touched code (during CoreHole parsing introduction).


electronicparsers/openmx/parser.py
```diff
@@ -463,26 +465,32 @@ def atom_index_dict(self) -> dict[str, list[int]]:
...
         Extract `Pseudopotential` and `CoreHole` (if present) from the atomic species definition.
         An explanation of the format can be found at https://www.openmx-square.org/openmx_man3.9/node32.html
         For an overview of all conventional potentials and partial atomic orbitals, see https://www.openmx-square.org/vps_pao2019/
         and https://www.openmx-square.org/vps_pao_core2019/ for core-holes.
         """
-        l_quantum = '[spdf]'
         remove_extension = lambda x: re.sub(r'(\.pao|\.vps)', '', x)
-        extract_method = lambda x: re.search(r'_([A-Z]+)19', x)
+        extract_method = lambda x: re.search(r'_([A-Z]+)\d\d', x)
+        extract_species = lambda x: re.search(rf'({element})_[A-Z]+\d\d', x)
```
ndaelman-hu: Minor nitpick

```suggestion
        extract_method = lambda x: re.search(r'_([A-Z]+)\d{2}', x)
        extract_species = lambda x: re.search(rf'({element})_[A-Z]+\d{2}', x)
```

electronicparsers/openmx/parser.py
```diff
@@ -463,26 +465,32 @@ def atom_index_dict(self) -> dict[str, list[int]]:
...
         #  extract_orbital = lambda x: re.search(rf'_(\d)({l_quantum})$', x)
-        extract_core_hole = lambda x: re.search(rf'_(\d)({l_quantum})_CH', x)
+        extract_core_hole = lambda x: re.search(rf'_(\d)({l_quantum})(_CH)?', x)
```
ndaelman-hu, line 483: Couldn't you at this point simply remove the appendix:

```suggestion
        extract_core_hole = lambda x: re.search(rf'_(\d)({l_quantum})', x)
```

electronicparsers/openmx/parser.py
```diff
@@ -463,26 +465,32 @@ def atom_index_dict(self) -> dict[str, list[int]]:
...
+            return None, definitions[0], 0, None
 
         definitions = deepcopy(definitions)
         try:
             definitions[1] = remove_extension(definitions[1])
             definitions[2] = remove_extension(definitions[2])
         except IndexError:
             self.logger.error(f'Species definition must be of length 3: {definitions}')
-            return None, None
+            return None, None, None, None
```
ndaelman-hu: If definitions are shorter than 3, wouldn't the top line also throw an error?

```python
if str(definitions[2]) == 'E'
```

I suggest having the first check be the length and then continue onward from there.

ndaelman-hu: Btw, with the new parsers, we would add a specific unit test for edge cases.
It would've pointed out this bug.

electronicparsers/openmx/parser.py
```diff
@@ -513,10 +521,10 @@ def parse_species(
                 )
             except KeyError:
                 self.logger.error(f'Unknown l-quantum symbol: {quantum_nums[1]}')
-                return pseudopotential, None
+                return pseudopotential, definitions[0], species, None
```
ndaelman-hu: Empty return values from `parse_species` were not properly captured. Thank you for covering these cases.

electronicparsers/openmx/parser.py
```diff
@@ -513,10 +521,10 @@ def parse_species(
...
 
             core_hole_flags = mainfile_parser.results.get('core_hole')
-            if core_hole_flags:
+            if core_hole_flags and quantum_nums[2] is not None:
```
ndaelman-hu, line 530: Why did add this check? I don't see `quantum_nums` being used below.

electronicparsers/openmx/parser.py
```diff
@@ -218,7 +220,7 @@ def convert_eigenvalues(string):
         ),
         Quantity(
             'core_hole',
-            r'<Definition\.of\.Core\.Hole([\s\S]+)Definition\.of\.Core\.Hole>',
+            r'<core\.hole\.state([\s\S]+)core\.hole\.state>',
```
ndaelman-hu, line 227: This is a pretty big change. Did the format get radically altered between versions or so? Should we maybe have both?

electronicparsers/openmx/parser.py
```diff
@@ -463,26 +465,32 @@ def atom_index_dict(self) -> dict[str, list[int]]:
...
     def parse_species(
         self, definitions: list[str]
-    ) -> tuple[Pseudopotential, Optional[CoreHole]]:
+    ) -> tuple[Pseudopotential, str, int, Optional[CoreHole]]:
         """
         Extract `Pseudopotential` and `CoreHole` (if present) from the atomic species definition.
         An explanation of the format can be found at https://www.openmx-square.org/openmx_man3.9/node32.html
         For an overview of all conventional potentials and partial atomic orbitals, see https://www.openmx-square.org/vps_pao2019/
         and https://www.openmx-square.org/vps_pao_core2019/ for core-holes.
         """
```
ndaelman-hu, line 478: `definitions` seem more structured than a random list. Maybe we should change it to a `dict[str, str]`?

ndaelman-hu (approved):
Thank you for the extension. The new code appears well inline with the existing structure, and minimal changes are overall preferable.

I have 2 slight worries:

1. You completely changed the format for matching the core holes. Could you explain why?
2. There is an edge case that is not properly captured atm.

electronicparsers/openmx/parser.py
ondracka: > If definitions are shorter than 3, wouldn't the top line also throw an error?
> 
> ```python
> if str(definitions[2]) == 'E'
> ```
> 
> I suggest having the first check be the length and then continue onward from there.

I did a quick check and OpenMX will error out if there are less than three columns in the species definitions, which means there will be no output file to parse. So this could happen only if the file gets corrupted. Is this something worth checking for?

ndaelman-hu: I think so. We would then log an error saying the file is corrupted. Question is whether anything else can then still be parsed.

electronicparsers/openmx/parser.py
ondracka: To be honest I don't know how this worked before, I cannot find any testcase where the Definition\.of\.Core\.Hole would be used. Even the docs show only core\.hole\.state works https://www.openmx-square.org/openmx_man3.9/node192.html . Can you maybe try to dig out some of the test cases you tested this against when introducing it initially?

electronicparsers/openmx/parser.py
ondracka: Right, this is kind of tricky, and indeed the current code is not 100% clear, but I'll try to explain what is going on and why I think this is needed.

What we are parsing here are the species definitions, like here:
```
<Definition.of.Atomic.Species
   C         C6.0_1s-s3p2d1      C_PBE19_1s
  Ti         Ti7.0-s3p2d1        Ti_PBE19
  ETi         Ti7.0-s3p2d1        E
  C1         C6.0_1s_CH-s3p2d1   C_PBE19_1s
Definition.of.Atomic.Species>
```

This is mostly just about what specific basis (second column) and pseudo (third column) are used for the specific species. The first line is the label. For example you would use C here for the C atom in the intial state calculation (there is technically no core-hole, even though we mark it as such with the 'intial' state). Than for the final state calculation one would just relabel the atom, but people don't usually bother to remove the species. 

Previously we incorrectly checked the basis for the corehole (for example we checked for 1s in C6.0_1s-s3p2d1) and also if there would be <core.hole.state in the input. That would however mark here both the C and C1 atom_parameters as final state. So we need specifically to check for the CH in the basis se "C6.0_1s_CH-s3p2d1" to be sure. If anything the check for `core_hole_flags` might be removable. The more complex way would be to check what atom index has core hole based on the core_hole_flags, than check what species corresponds to the atom index and than only mark the final state. But I think like this its simpler.

electronicparsers/openmx/parser.py
ondracka: This `_CH` is exactly what determines if this is the final or initial state (previously it was based on the `core_hole_flags` but that was not working properly, see below.)

electronicparsers/openmx/parser.py
ondracka: OK, will fix.

electronicparsers/openmx/parser.py
ondracka: Well yeah, the first column is always the label, second is the basis and third is the pseudo. I can update it.

electronicparsers/openmx/parser.py
ndaelman-hu: ow, okay, and you want to capture it as the 3rd group. all good then

electronicparsers/openmx/parser.py
ondracka: Actually, this broke it, since this is f-string. so `\d{{2}}` it is...

electronicparsers/openmx/parser.py
ondracka: Updated it to a dict (using str_operation at TextParser level)

electronicparsers/openmx/parser.py
ondracka: I moved it into the try block, so if anything is missing we bail out properly.
That said, the already at the TextParser level we expect three columns, so if we have less we never get this far IMO.

ondracka:
@ndaelman-hu I think I've incorporated all the required changes. Is this good to go?

EDIT: sorry, I've missed the approval. Will merge.

electronicparsers/openmx/parser.py
ndaelman-hu: I'm taking a look now, sorry for not answering here yet. From your provided documentation, I am pretty convinced by the soundness. This doesn't have to stall a merge.

ndaelman-hu: I've looked into this and you're right - I can't find any tests or documentation backing up the original `<Definition.of.Core.Hole>` pattern. I was likely just imitating other OpenMX fields like `<Definition.of.Atomic.Species>` without checking actual output files.

Your correction to `<core.hole.state>` is correct according to the docs. No need to support both patterns since the old one was never right to begin with. Thanks for catching this!

PR #331, opened 2026-04-14
Fix VASP time calculation dimensionality error

ndaelman-hu:
## Summary
Fixes a `pint.DimensionalityError` when parsing VASP vasprun.xml files with time calculation data.

## Problem
The parser was failing with:
```
pint.errors.DimensionalityError: Cannot convert from 'second' to 'dimensionless'
```

This occurred due to two issues:

### Issue 1: Missing units on time_calculation
The `time_val` from the XML parser is a plain number (float) without units. The original code assigned it directly:
```python
sec_scc.time_calculation = time_val  # No units!
```

When adding to `time_initial` (which has seconds units):
```python
sec_scc.time_physical = time_initial + sec_scc.time_calculation
# Error: Cannot add Quantity(seconds) + dimensionless float
```

Pint cannot add values with incompatible dimensionality (seconds vs dimensionless).

### Issue 2: None propagation in time_initial
`time_initial` could be `None` when previous calculations didn't have `time_physical` set, causing arithmetic failures when trying to add `None + Quantity`.

## Solution
- Add explicit units (`ureg.s`) when setting `time_calculation` values to ensure they're always proper quantities
- Store time quantity in a local variable before assignment for clarity
- Add `None` checks for `time_initial` with fallback to `0 * ureg.s`
- Add `None` checks for `time_physical` before using in calculations

## Testing
- ✅ All 23 VASP parser tests pass, including `test_malformed_time_entries`
- ✅ Successfully parses previously failing vasprun.xml files

### DIFF
--- a/electronicparsers/abacus/parser.py
+++ b/electronicparsers/abacus/parser.py
@@ -1184,13 +1184,13 @@ class ABACUSOutParser(TextParser):
             Quantity(
                 'program_version',
                 r'Version:\s*(.*)\n',
-                str_operation=lambda x: ''.join(x),
+                str_operation=''.join,
             ),
             Quantity('nproc', r'Processor Number is\s*(\d+)\n', dtype=int),
             Quantity(
                 'start_date_time',
                 r'Start Time is\s*(.*)\n',
-                str_operation=lambda x: ''.join(x),
+                str_operation=''.join,
             ),
             Quantity(
                 'input_filename',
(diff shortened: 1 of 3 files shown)


ladinesa (approved):

ndaelman-hu:
Also confirmed the fix in the GUI now, with the correct time extraction.
Both local parsing and the GUI confirm that SCF time stamps are only set when legible.

Screenshot from 2026-04-15 12-13-39
Screenshot from 2026-04-15 12-12-46

PR #332, opened 2026-04-27
Fix exciting bse parsing

ladinesa:
Address new issues with bse parsing in exciting. 
[issues.pdf](https://github.com/user-attachments/files/27129974/issues.pdf)


ladinesa:
@kubanmar please test this in your oasis.

